PR #1, opened 2018-09-15
Avoid unnecessary/incorrect let-bindings

vermiculus:
The `let-alist` macro does not handle nested calls well: every member within the inner call will be pulled up extraneously to the outer call and bound there as well.  By moving the inner call to a new function, we hide the inner call from the macro expander.

---

I know you're still actively working on this and may wish to address this another way by now, but I did want to point out this pattern as something to avoid.

### DIFF
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -161,20 +161,23 @@ repositories.
         (oset issue locked-p   .locked)
         (oset issue milestone  .milestone)
         (oset issue body       (forge--sanitize-string .body))
-        (dolist (c .comments)
-          (let-alist c
-            (closql-insert
-             (forge-db)
-             (forge-issue-post
-              :id      (forge--object-id issue-id .databaseId)
-              :issue   issue-id
-              :number  .databaseId
-              :author  .author.login
-              :created .createdAt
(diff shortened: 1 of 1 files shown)


tarsius:
That doesn't appear to be true.

```lisp
(let (var)
  (let-alist var
    .outer
    (let-alist var
      .inner))
```

expands to

```lisp
(let (var)
  (let
      ((alist var))
    (let
        ((\.outer
          (cdr
           (assq 'outer alist))))
      \.outer
      (let
          ((alist var))
        (let
            ((\.inner
              (cdr
               (assq 'inner alist))))
          \.inner)))))
```

(I did this with `macrostep-expand`.)

If it were true, then I would expect the byte-compiler to complain.

Also the `let-alist.el` commentary warns

> If you nest `let-alist` invocations, the inner one can't access the variables of the outer one.

If what you say were true, I would expect it to warn about that too. The documentation does not say that one should not nest `let-alist` -- it merely warns about the above limitation.

vermiculus:
Alas, it seems updates were made since I last looked at the source code (or perhaps I've remembered incorrectly, plain and simple). Sorry for the bad assumptions!

PR #11, opened 2018-12-20
Fix/gitlab url schema

dsdshcym:
This PR does 2 things:

1. Add `gitlab@` as a valid prefix to `forge--url-regexp`
2. Support multi-level groups path for gitlab. But the fix for this is a dirty hack. i need some guidance about how to clean it up

Thanks!

### DIFF
--- a/lisp/forge-core.el
+++ b/lisp/forge-core.el
@@ -220,9 +220,9 @@ at that time."
 
 (cl-defmethod forge--split-url-path
   ((_class (subclass forge-repository)) path)
-  (and (string-match "\\`\\([^/]+\\)/\\([^/]+?\\)\\'" path)
+  (and (string-match "\\`\\([^/]+\\)/\\(.+?\\)\\'" path)
        (list (match-string 1 path)
-             (match-string 2 path))))
+             (replace-regexp-in-string "/" "%2F" (match-string 2 path)))))
 
 (cl-defmethod forge--split-url-path
   ((_class (subclass forge-noapi-repository)) path)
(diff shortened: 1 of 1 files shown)


tarsius:
> add gitlab@ as a valid prefix to forge--url-regexp

Could you please point me to the documentation?  In the web interface it only says:

> Clone with SSH 
>   git@gitlab.com:magit/testibus.git
> Clone with HTTPS
>   https://gitlab.com/magit/testibus.git

tarsius:
> Support multi-level groups path for gitlab

I have decided to make the subgroup part of the owner instead of the
repository name, because that corresponds to how Gitlab sees it.

You'll also have to update `ghub` because I had to do the same there.

> Add gitlab@ as a valid prefix to forge--url-regexp

Well, it can be anything really, so we'll need an option.  Still
struggling with the name and doc-string though

``` emacs-lisp
(defcustom forge-ssh-users
  '("git" "gitlab" "teahub")
  "Additional ssh users used on forge hosts.

Given an url like \"git@gitlab.com\" or \"ssh://git@gitlab.com\"
\"git\" is the ssh user on the forge server.  When connecting to
a repository hosted on a forge, then that user does not actually
correspond to a user account belonging to a human.  All humans
that use the ssh protocol authenticate as the same user and the
provided public key is used to actually tell the different humans
apart.

The major forge hosts all use \"git\", but some private and less
popular public hosts may use something else, in which case you
have to add the name here."
  :package-version '(forge . "0.2.0")
  :group 'forge
  :type '(repeat string))
```

Could you please help me with that?

dsdshcym:
@tarsius Of course.

This weekend, I'll try to find the documentation for `gitlab@` (or it might be the username my company is using) and add the `forge-ssh-users` (with a better name maybe) option to `forge`.

tarsius:
#35 is related.

tarsius:
We might just want to allow *any* username as suggested there instead of adding an option. And we have to adjust the regexp in two places, regardless of whether we use an option or not.

PR #31, opened 2018-12-21
Update the casing of Github to GitHub

PuercoWork:
GitHub is a proper name and has special casing rules

(Sorry for the nit, Thanks for the all the software you write)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Work with Git forges from the comfort of Magit
 ==============================================
 
-Work with Git forges, such as Github and Gitlab, from the comfort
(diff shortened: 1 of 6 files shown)


tarsius:
Likewise for GitLab, and it's also not Srht but sr.ht.

I have long ago decided to normalize such names to follow the
conventions of the English language (as I believe to know them).  I
don't actually care much about it, but I also see no reason to change
this.  Especially because if I did change it, then I would want do it
properly, i.e. in all of my packages and also not just changing it in
one case, GitHub, but not the other, Gitlab.  That would be much more
work, and fairly useless.  At least now it is consistent (minus some
contributed strings I did not bother to change).

PR #35, opened 2018-12-21
Correct username handling for ssh URLs in `forge--url-regexp'

snogge:
The git-clone(1) man page states that ssh git urls should follow

ssh://[user@]host.xz[:port]/path/to/repo.git/

Make the username part optional and allow usernames other than `git'.
The supplied regexp should match all allowed github usernames, but is
not very strict.


tarsius:
#11 is related.

Issue #41, opened 2018-12-25
Support for inserting images in PRs and issues


dusek:
It would be cool if one could insert images in PRs and issues. E.g. for my current PR, I am inserting a few screenshots of a new interface to give the reviewer better idea what the proposed changes lead to. (Screenshots can similarly be useful when reporting issues). I have two options now:

1. Avoid using forge and create the PR directly on GitHub.
2. Create the PR in forge but with placeholder text instead of the screenshots, then open the PR on GitHub and edit it and add the actual screenshots (this has the disadvantage of notifying the reviewer by email only after creating the PR without screenshots, missing the opportunity to make first impression; the subsequent edit on GitHub will not be notified to the reviewer and the screenshots might get unnoticed for quite some time).

I would love for there to be a third option - I can create the PR with the screenshot already directly in forge 😉.

Same for videos, see  #49.

vermiculus:
Is there a standard emacs convention for working with images in documents?

tarsius:
Lets start with displaying the images that are already there. Look at `markdown-display-inline-images` for inspiration.

vermiculus:
I think displaying images would be best handled by markdown-mode itself. I'm investigating one approach to that by altering the function you mention. Right now as released, it only supports showing local files – but you've gotta download the image eventually. I figure that function can be augmented to optionally download remote images it finds.

vermiculus:
This definition of `markdown-display-inline-images` *does* work to display images, *but* the image is being removed as it's inserted into the buffer. Obviously it's very rough around the edges and getting the image is an absolute hack (and we'd want some kind of caching, probably), but it works as a proof-of-concept.

```elisp
(defun markdown-display-inline-images ()
  "Add inline image overlays to image links in the buffer.
This can be toggled with `markdown-toggle-inline-images'
or \\[markdown-toggle-inline-images]."
  (interactive)
  (unless (display-images-p)
    (error "Cannot show images"))
  (save-excursion
    (save-restriction
      (widen)
      (goto-char (point-min))
      (while (re-search-forward markdown-regex-link-inline nil t)
        (let ((start (match-beginning 0))
              (imagep (match-beginning 1))
              (end (match-end 0))
              (file (match-string-no-properties 6)))
          (when (and imagep
                     (not (zerop (length file))))
            (unless (file-exists-p file)
              ;; try to download it
              (cond
               ((string-match-p "\\`https?://.+" file)
                (let ((url file))
[...]

PR #47, opened 2018-12-29
Display inline images in topics

vermiculus:
Implement the view-parts of #41.

This is still something of a proof-of-concept and is dependent on jrblevin/markdown-mode#378. Make sure you set `markdown-display-remote-images` to `t` (it could be let-bound to a Forge-customization here) and that all appropriate protocols have been added to `markdown-remote-image-protocols`.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -334,7 +334,10 @@ identifier."
             (add-face-text-property 0 (length heading)
                                     'magit-diff-hunk-heading t heading)
             (magit-insert-heading heading))
-          (insert (forge--fontify-markdown body) "\n\n"))))))
+          (insert (forge--fontify-markdown body) "\n\n"))))
+    (when (fboundp 'markdown-display-inline-images)
+      (let ((markdown-display-remote-images t))
+        (markdown-display-inline-images)))))
 
 (defvar forge-topic-title-section-map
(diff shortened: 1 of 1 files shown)


vermiculus:
From https://github.com/magit/forge/issues/41#issuecomment-450487363
> Lets see if using `about` actually works:

Funnily enough, this is causing an error. Don't know why EXTERN is being caught here – probably being caught as real markdown even though it's in a code-block. I think it might have something to do with hiding markup, but both `markdown-hide-markup` and `markdown-hide-markup-in-view-modes` are nil.

Your note about overlays not being copied may be relevant here.

tarsius:
Not sure why I haven't done anything about this for so long. It seems to work well for the most part.

> (it could be let-bound to a Forge-customization here) 

Done that in a fixup commit.

> Funnily enough, this is causing an error.

But that's something that has to be addressed in `markdown-mode`, right? It would be nice if this would continue after failing to insert one image. (Or at least that's something worth considering).

Unless you see a reason to delay this, I'll merge it.

vermiculus:
Testing this since it's been a while… :-)

image

vermiculus:
Once I took out `about` from my post above, still works fine :)

Looks like `markdown-regex-link-inline` needs to be adjusted to avoid matching `` `blocks` ``.

Issue #45, opened 2018-12-28
Quoted blocks removed when visiting topics


PuercoPop:
In particular I've encountered this behaviour in this Pull Request

https://github.com/stumpwm/stumpwm/pull/544#issuecomment-449791330

![image](https://user-images.githubusercontent.com/387111/50526343-66194f00-0aaf-11e9-89e7-b0271d3815a2.png)

Note that there is a line with one empty space where the quoted block should be. Also I verified in the SQLite database that the quoted blocks are present

vermiculus:
This might be a gfm-like setting; did you check any variables that might be hiding markup? (My wife is hoarding my laptop right now but I *think* it's labeled as some sort of 'clean' mode.)

PuercoPop:
> This might be a gfm-like setting; did you check any variables that might be hiding markup? (My wife is hoarding my laptop right now but I think it's labeled as some sort of 'clean' mode.)

I can't find anything that might be doing so, but you've given me an idea. It may be that markdown-mode may be interpreting the content as the lang of the code-block as the whole quoted text is one line with triple backticks

PuercoPop:
@vermiculus Yes, I quote with `>` but I don't control how other people use markup. Thanks for the pointer of `markdown-hide-markup-in-view-modes` but that didn't do the trick. Although I suspect the 'bug' is more likely to be in markdown-mode, I'll look into it. Thanks for the pointers again!

vermiculus:
I find it unlikely that markdown-mode will ever do what you're asking it to do (which is to effectively ignore the most basic syntax rules of markdown!), but you might be able to hack the regular expressions markdown-mode uses to recognize quote blocks. You'll have better luck convincing others to use markdown correctly, I think :-)

It is odd though that from getting the fontified text in `forge--fontify-markdown` to inserting it in `forge-topic-refresh-buffer`, we seem to loose the content in ```` ```triple quotes``` ````.

tarsius:
`one`

``two``

```tree```

tarsius:
`markdown-mode` sets `invisible` to `markdown-markup` for all of

    ```text```

If the delimiters are just one or two backquotes, then it doesn't do that. At least `gfm-mode` shouldn't treat three backquotes the same as two because three opening *and* closing backquotes on the same line is valid gfm syntax and in that case the `text` part is not part of the markup its the content, unlike in

    ```language
    content
    ```

where `language` is part of the markup.

----

[...]

PR #53, opened 2018-12-30
Bind RET to follow Markdown links in issue buffers

Miciah:
* `lisp/forge-topic.el` (`forge-topic-mode-map`): Remap `magit-visit-thing` to `markdown-follow-link-at-point`.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -205,6 +205,7 @@ The following %-sequences are supported:
     (define-key map (kbd "C-c C-n") 'forge-create-post)
     (define-key map (kbd "C-c C-r") 'forge-create-post)
     (define-key map [remap magit-browse-thing] 'forge-browse-topic)
+    (define-key map [remap magit-visit-thing] 'markdown-follow-link-at-point)
     map))
 
 (define-derived-mode forge-topic-mode magit-mode "View Topic"


tarsius:
Yes, we cannot just bind this in `forge-topic-mode-map`. Doing it properly will probably be quite a bit of work. This also came up in #45.

Miciah:
I almost went with a wrapper instead of binding `markdown-follow-link-at-point` directly, something like the following:
```emacs-lisp
(defun forge-issue-visit-thing ()
  "Visit the current link using a browser."
  (interactive)
  (cond ((markdown-link-p)
         (markdown-follow-link-at-point))
        ((markdown-wiki-link-p)
         (markdown-follow-wiki-link-at-point))
        ;; TODO Handle @username and #123.
        (t (user-error "Nothing to follow at point"))))
```
In the end I went with binding`markdown-follow-link-at-point` directly because refactoring is the easy part (compared to implementing the TODO), but I can amend my change to use the above function; please let me know if that would be an acceptable starting point.

Another option would be to scan for properties or regexes for Markdown links and using `make-text-button` or similar to linkify them.  Please let me know if you would be interested in that approach.

tarsius:
I'm undecided. Reopening to keep track of this.

tarsius:
> Another option would be to scan for properties or regexes for Markdown links and using `make-text-button` or similar to linkify them.  Please let me know if you would be interested in that approach.

I think we should use this approach.

PR #74, opened 2019-01-16
Add "Forge" to magit-dispatch-popup

MatthewFluet:
For users who forget the `forge-dispatch` key, it is helpful to
include it in `magit-dispatch-popup`.


tarsius:
Thanks!

PR #77, opened 2019-01-17
Bind forge-list-visit-* in forge-*-list-mode-map w/ magit convention

MatthewFluet:
Follow the convention established by magit for `magit-visit-thing`;
this allows a terminal client to visit topics in a topic list mode
with `RET`, just as commits, topics, etc. can be visited with `RET` in
a magit-mode.

See magit/magit@f629795 for a description of the jkl feature.

### DIFF
--- a/lisp/forge-list.el
+++ b/lisp/forge-list.el
@@ -56,7 +56,10 @@
 (defvar forge-issue-list-mode-map
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map forge-topic-list-mode-map)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks a lot for taking the time to learn why I used `[return]` instead of just "fixing" the issue.  But I think using `cond` is overkill in this particular case and I went with just always binding both keys.

MatthewFluet:
Fair enough.  Indeed, I had added the `(kbd "RET")` binding via my `.emacs`, without removing the `[return]` binding, with no ill-effects.

PR #95, opened 2019-01-28
Provide context when creating posts

vermiculus:
Often times I get confused as to which repository I'm working in.  This patch uses the header-line (as Magithub does) to contextualize the buffer.

### DIFF
--- a/lisp/forge-post.el
+++ b/lisp/forge-post.el
@@ -109,7 +109,7 @@
 (defvar-local forge--cancel-post-function nil)
 (defvar-local forge--pre-post-buffer nil)
 
-(defun forge--prepare-post-buffer (filename)
+(defun forge--prepare-post-buffer (filename &optional header)
   (let ((file (magit-git-dir
                (convert-standard-filename
                 (concat "magit/posts/" filename)))))
@@ -119,6 +119,8 @@
           (buf (find-file-noselect file)))
       (with-current-buffer buf
         (forge-post-mode)
+        (when header
+          (magit-set-header-line-format header))
         (setq forge--pre-post-buffer prevbuf)
         (when resume
           (forge--display-post-buffer buf)
--- a/lisp/forge-post.el
+++ b/lisp/forge-post.el
@@ -51,11 +51,11 @@
 ;;; Utilities
 
 (cl-defmethod forge-get-url ((post forge-post))
-  (forge--format-url post (let ((topic (forge-get-parent post)))
-                            (cond ((forge--childp topic 'forge-issue)
-                                   'issue-post-url-format)
-                                  ((forge--childp topic 'forge-pullreq)
-                                   'pullreq-post-url-format)))))
+  (forge--format post (let ((topic (forge-get-parent post)))
+                        (cond ((forge--childp topic 'forge-issue)
+                               'issue-post-url-format)
+                              ((forge--childp topic 'forge-pullreq)
+                               'pullreq-post-url-format)))))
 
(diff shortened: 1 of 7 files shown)


lisp/forge-post.el
```diff
@@ -98,7 +98,7 @@
 (defvar-local forge--cancel-post-function nil)
 (defvar-local forge--pre-post-buffer nil)
 
-(cl-defun forge--prepare-post-buffer (filename repo)
+(cl-defun forge--prepare-post-buffer (filename repo &key header)
```
tarsius: Do you expect to add more arguments or why did you choose to use a keyword argument?

What do you think about making `repo` the first argument?

lisp/forge-post.el
```diff
@@ -88,7 +88,8 @@
     (define-key map (kbd "C-c C-k") 'forge-post-cancel)
     map))
 
-(define-derived-mode forge-post-mode gfm-mode "Forge-Post" "")
+(define-derived-mode forge-post-mode gfm-mode "Forge-Post" ""
+  nil)
```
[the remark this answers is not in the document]
vermiculus: Ah, sorry; I can revert this. I originally had the header-line-format stuff in here instead of prepare-post-buffer.

lisp/forge-post.el
vermiculus: `repo` being the first argument: I'm a fan; makes sense. I've made this update.

I expect more arguments might be needed over time, but this might be a coding style precaution carried over from work (where function signatures – even private ones – are rarely changed). I can promote it to a formal, optional argument, but I do think keys provide some readability benefits.

tarsius:
I have pushed an alternative approach to the `post-header` branch. Obviously it is not done; most importantly `forge--format-url` would have to be renamed. Please have a look anyway.

vermiculus:
I love it! Using `forge--format-url` (`forge--format`?) for this purpose is a more powerful abstraction. Do you mind if I adopt that implementation here in this PR to keep the discussions linked up?

tarsius:
No. And I have just done that.

vermiculus:
I've renamed `forge--format-url` to `forge--format` as I mentioned above.  Assuming it's all still compiling as before (this branch doesn't have the changes from #100 so it's difficult to test), these changes look good!

Do you generally create a merge commit or do you rebase onto `master`?

tarsius:
Big fan of the rebase :stuck_out_tongue_winking_eye:

lisp/forge-repo.el
```diff
@@ -231,9 +231,11 @@ forges and hosts.  "
                       :limit 1]
                      table (oref repo id)))))
 
-(cl-defmethod forge--format-url ((repo forge-repository) slot &optional spec)
+(cl-defmethod forge--format ((repo forge-repository) slot &optional spec)
   (format-spec
-   (eieio-oref repo slot)
+   (if (symbolp slot)
+       (eieio-oref repo slot)
+     slot)
```
tarsius: Since we use it like that we should rename the argument to `format-or-slot`

vermiculus:
I've renamed `slot` to `format-or-slot` and rebased onto current `master`.

tarsius:
Could you please rebase again?  There's gonna be conflicts this time around.

tarsius:
My strategy to do that would be to first recreate the second commit from scratch on `master` and then squash the two pr commits before rebasing the result.

vermiculus:
Rebased. The only real conflict I could find (not detected automatically) was some straggling uses of `forge--format-url`. I tested the compile with #100 with no issues.

Issue #88, opened 2019-01-24
Hide closed PRs in PR list


Linuus:
In the magit status buffer there is a "Pull requests" section which lists the pull requests. The closed PRs are highlighted differently.

Is it possible to not show closed PRs at all in this list?

vermiculus:
Related #42 #79

tarsius:
You can customize `forge-topic-list-limit` to do that. Well almost, closed but unread topics are shown anyway.

Linuus:
@tarsius Ah nice, so setting it to something like `(60 . 0)` should do it.

Thanks!

Issue #79, opened 2019-01-18
Custom magit-section: assigned pull requests


alexhenning:
Not an issue, but I thought it might be helpful for other people. This is an example how I created a custom `magit-section` for pull-requests assigned to me, so I can quickly focus in on the pull-requests that have been assigned to me:

```
Magit status buffer
...
Assigned pull requests (1)
#486 A PR that needs my attention

Pull requests (24)
#485 Awesome new PR with fancy features
...
```

Actual snippet:

```elisp
(defun alexhenning/get-assigned-pullreqs (repo)
  (let* ((raw-ids (forge-sql [:select pullreq:id :from pullreq
                                      :inner :join pullreq_assignee :on (= pullreq_assignee:pullreq pullreq:id)
                                      :inner :join assignee :on (= pullreq_assignee:id assignee:id)
                                      :where (and (= pullreq:repository $s1) (isnull pullreq:closed) (= assignee:login $s2))
                                      :order-by [(desc updated)]]
                             (oref repo id) "alexhenning"))
         (ids (apply 'vector (mapcar 'car raw-ids)))
         (pullreqs (forge-sql [:select * :from 'pullreq
                                       :where id :in $v1
                                       :order-by [(desc updated)]]
                       ids)))
    (cl-sort (mapcar (lambda (row)
                       (closql--remake-instance 'forge-pullreq (forge-db) row))
                     pullreqs)
             (cdr forge-topic-list-order)
             :key (lambda (it) (eieio-oref it (car forge-topic-list-order))))))

(defun alexhenning/insert-assigned-pullreqs ()
  (when-let ((repo (forge-get-repository nil))
             (- (not (oref repo sparse-p)))
             (pullreqs (alexhenning/get-assigned-pullreqs (forge-get-repository nil))))
    (magit-insert-section (pullreqs nil t)
      (magit-insert-heading
        (format "%s (%s)"
                (propertize "Assigned pull requests" 'face 'magit-section-heading)
                (length pullreqs)))
      (magit-insert-section-body
        (let ((width (length (number-to-string (oref (car pullreqs) number))))
              (prefix (forge--topic-type-prefix (car pullreqs))))
          (dolist (pullreq pullreqs)
            (forge-insert-pullreq pullreq width prefix)))
        (insert ?\n)))))

(magit-add-section-hook 'magit-status-sections-hook 'alexhenning/insert-assigned-pullreqs 'forge-insert-pullreqs)
```

[...]

PR #90, opened 2019-01-26
Generalize browse, visit commands and introduce copy-URL command

vermiculus:
Fix #85.

Find in this PR a series of patches that generalizes the existing browse/visit commands and introduce a new command, `forge-copy-url-at-point-as-kill`. I recommend it be bound as `C-c M-w` next to `magit-browse-thing`.

A new method `forge-get-url` is introduced to simplify the function above.  Implementations are provided for issues, pullreqs, posts (to handle comments), repositories, and notifications.  The latter two are unused at this time, but exist for completeness.

At that point, it made sense to redefine `forge-browse-*` in terms of `forge-get-url`.  This is done using another new method, `forge-browse`, where an implementation is added for posts.  The new implementation is decidedly simpler, but perhaps that means I'm missing something here.

And just because the `forge-visit-*` functions then stuck out like sore thumbs, a new method `forge-visit` was provided for those as well -- with a single implementation for posts.  This reduced duplication quite a bit.

Let me know if there are any changes I should make.

I'll note that `forge--remote` was added and used in an incorrect method implementation of `forge-get-url` for `forge-repository` objects. It's only used once again, but I get the feeling it's good to separate out. Since it's not strictly necessary, though, I'll remove it if you so request.

---

I'm thoroughly enjoying working with Forge -- I'm even starting work on a custom `forge-repository` for our home-grown bug-tracking system at work.  It's challenging on that end, but I'm finding Forge well-designed to handle what I throw at it (at least most of it -- our code base is in SVN, so that presents its own troubles).  I may look for help/guidance on that in the future, but I wanted to take this opportunity to say that Forge is a job well done -- I look forward to its continued growth :smile:

PR #97, opened 2019-01-31
Generalize topic list insertion

vermiculus:
Address issues raised in #79 and #88.

Find below patches that generalize topic and topic-list insertion.  The stars of the show are `forge-insert-topics`, `forge-insert-topic`, and `forge--insert-topic-contents`.

`forge-insert-topics` inserts the section containing the individual topic-sections.  This is the function to use if you want a list of topics.  (I call this the list-section.)

`forge-insert-topic` inserts the section for an individual topic.  (I call this the topic-section.)  This wouldn't be necessary save for topic insertion in notification lists; perhaps there is a better way, but I couldn't find one.

`forge--insert-topic-contents` is the meat of the section -- both the title line and the section contents in the case of pullreqs.  This method uses `forge--format-topic-id` to handle forge-specific semantics (i.e., Gitlab using `!` to reference pull requests) and centralizes the faces used for unread/closed/open topics as well as label insertion after the title.  To effect the listing of commits for pullreqs, an `:after` implementation just for pullreqs is used to insert the heading 'marker' (I call it this, but I don't actually know the implementation) and the actual commit log as before.

---

I hope the above patches prove worthwhile in their own right by reducing duplication, but I think they'll also be very convenient for implementing either `defcustom`-based configuration of any single list-section or defining separate functions for multiple different views of issues/pullreqs within the same status buffer.

The following code is proof-of-concept only; it clearly doesn't function right now.

### Using `defcustom`

``` emacs-lisp
(defun forge-insert-pullreqs ()
  (when-let ((repo (forge-get-repository nil))
             (- (not (oref repo sparse-p))))
    (let ((pullreqs (pcase forge-pullreq-list-configuration
                      (`(by-author . ,author)
                       ;; etc.
                       )
                      (`(by-assignee . ,assignee)
                       ;; etc.
                       )
                      (_
                       ;; using separate configuration in `forge-topic-list-limit'.
                       ;; perhaps it would be good to assume it here.
                       (forge-list-recent-topics repo 'pullreq)))))
      (forge-insert-topics "Pull requests" pullreqs))))
```

### Using separate functions
At its simplest, you can simply copy/paste `forge-insert-pullreqs` using a different source list for the pull requests.  You could get arbitrarily fancy with it, though.

``` emacs-lisp
(defun forge-insert-pullreqs--create (title &optional config)
  (let ((pullreqs-source (pcase config
                           (`(by-author . ,author)
                            ;; etc.
                            )
                           (`(by-assignee . ,assignee)
                            ;; etc.
                            )
                           (_
                            ;; using separate configuration in `forge-topic-list-limit'.
                            ;; perhaps it would be good to assume it here.
                            (lambda () (forge-list-recent-topics repo 'pullreq))))))
    (lambda ()
      (when-let ((repo (forge-get-repository nil))
                 (- (not (oref repo sparse-p))))
        (forge-insert-topics title (funcall pullreqs-source))))))

;; or using a `defmacro' or whatever you'd fancy
(defalias forge-insert-pullreqs-assigned-to-me
  (forge-insert-pullreqs--create
   "PRs assigned to me"
   ;; using hypothetical `forge-current-user'
   `(by-assignee . ,(forge-current-user))))

(defalias forge-insert-pullreqs-created-by-me
  (forge-insert-pullreqs--create
   "My PRs"
   ;; using hypothetical `forge-current-user'
   `(by-author . ,(forge-current-user))))

(defalias forge-insert-pullreqs
  ;; default behavior
  (forge-insert-pullreqs--create "Pull requests"))
```

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -226,6 +226,58 @@ The following %-sequences are supported:
       (replace-regexp-in-string "\r\n" "\n" string t t)
     ""))
 
+(defun forge-insert-topics (heading topics)
+  "Under a new section with HEADING, insert TOPICS."
+  (when topics
+    (let ((width (length (number-to-string (oref (car topics) number))))
+          list-section-type topic-section-type)
+      (cond ((forge--childp (car topics) 'forge-issue)
+             (setq list-section-type  'issues)
+             (setq topic-section-type 'issue))
+            ((forge--childp (car topics) 'forge-pullreq)
+             (setq list-section-type  'pullreqs)
+             (setq topic-section-type 'pullreq)))
+      (magit-insert-section ((eval list-section-type) nil t)
+        (magit-insert-heading
+          (format "%s (%s)"
+                  (propertize heading 'face 'magit-section-heading)
+                  (length topics)))
+        (magit-insert-section-body
+          (dolist (topic topics)
+            (forge-insert-topic topic topic-section-type width))
+          (insert ?\n))))))
+
+(defun forge-insert-topic (topic &optional topic-section-type width)
+  "Insert TOPIC as a new section.
+If TOPIC-SECTION-TYPE is provided, it is the section type to use.
+If WIDTH is provided, it is a fixed width to use for the topic
+identifier."
+  (unless topic-section-type
+    (setq topic-section-type
+          (cond ((forge--childp topic 'forge-issue) 'issue)
+                ((forge--childp topic 'forge-pullreq) 'pullreq))))
+  (magit-insert-section ((eval topic-section-type) topic t)
+    (forge--insert-topic-contents topic width)))
+
+(cl-defmethod forge--format-topic-id ((topic forge-topic))
+  (propertize (format "#%s" (oref topic number)) 'face 'magit-dimmed))
+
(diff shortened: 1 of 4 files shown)


lisp/forge-topic.el
```diff
@@ -216,6 +216,63 @@ The following %-sequences are supported:
       (replace-regexp-in-string "\r\n" "\n" string t t)
     ""))
 
+(cl-defmethod forge--prepare-post-edit-buffer-for ((topic forge-topic))
```
vermiculus: This is not supposed to be in this commit; I'll remove when I get home tonight.

tarsius:
I have done the initial review in form of a few fixup commits.

vermiculus:
For types not known at compile-time, would be an option to [eval-and-compile](https://github.com/vermiculus/ghub-plus/blob/51ebffe549286b3c0b0565a373f44f4d64fc57af/ghub%2B.el#L37) those types?

tarsius:
That wouldn't keep the compiler from complaining.

vermiculus:
If `forge--childcase` from #90 is to your liking, I'll use it here.

PR #100, opened 2019-02-14
Use EMake for the build

vermiculus:
Key advantages:

- There is no longer an expected structure above the root directory of this project.  (I can compile with `make clean lisp` now... #90)

- The dependency list is no longer maintained in two places (`default.mk` and `*-pkg.el`)

Note some setup will have to be done in config.mk for packages that are available from above the root directory of this project; make sure to add the necessary load-path code to `EMACS_ARGS`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@
 /docs/*.pdf
 /docs/dir
 /docs/forge/
+/.emake/
 /lisp/forge-autoloads.el
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ help:
 	$(info make publish      - publish snapshot manuals)
 	$(info make release      - publish release manuals)
 	$(info make clean        - remove most generated files)
+	$(info make run          - run Emacs with minimal configuration)
+	$(info make dev-run      - like 'make run', but use local packages when available)
+	$(info make deps         - install dependencies)
 	@printf "\n"
 
 lisp:
@@ -50,3 +53,11 @@ clean:
 	@printf "Cleaning...\n"
 	@$(MAKE) -C lisp clean
 	@$(MAKE) -C docs clean
+	@$(MAKE) -f emake.mk clean
+
+dev-run:
+	@$(MAKE) -f emake.mk dev-run
+run:
+	@$(MAKE) -f emake.mk run
+deps:
+	@$(MAKE) -f emake.mk $(EMAKE_WORKDIR)/elpa
--- a/default.mk
+++ b/default.mk
@@ -19,24 +19,29 @@ ELS  += $(PKG)-semi.el
 ELS  += $(PKG)-commands.el
 ELS  += $(PKG)-list.el
 ELCS  = $(ELS:.el=.elc)
-
-DEPS  = closql
-DEPS += dash
-DEPS += hydra # for lv.el
-DEPS += emacsql
-DEPS += ghub
-DEPS += graphql
-DEPS += magit/lisp
-DEPS += markdown-mode
-DEPS += transient/lisp
-DEPS += treepy
-DEPS += with-editor
+PACKAGE_LISP = $(addprefix lisp/,$(ELS))
 
 EMACS      ?= emacs
 EMACS_ARGS ?=
 
-LOAD_PATH  ?= $(addprefix -L ../../,$(DEPS))
-LOAD_PATH  += -L .
+CURL       ?= curl -fsSkL --retry 9 --retry-delay 9
+
+EMAKE_SHA1 ?= e2db14acc21fdf9e0c76377c5165149562c1ed88
+
+EMAKE_WORKDIR  ?= .emake
+EMAKE_LOGLEVEL ?= INFO
+
+EMAKE_ENV += PACKAGE_FILE="$(PKG)-pkg.el"
+EMAKE_ENV += PACKAGE_LISP="$(PACKAGE_LISP)"
+EMAKE_ENV += PACKAGE_ARCHIVES="melpa"
+EMAKE_ENV += EMAKE_WORKDIR="$(EMAKE_WORKDIR)"
+EMAKE_ENV += EMAKE_LOGLEVEL="$(EMAKE_LOGLEVEL)"
+
+EMAKE_BASE ?= $(EMAKE_ENV) $(EMACS) -Q $(EMACS_ARGS) \
+	-L '$(abspath lisp)' \
+	-l '$(EMAKE_WORKDIR)/emake.el' \
+	--eval '(setq enable-dir-local-variables nil)'
+EMAKE ?= $(EMAKE_BASE) --eval "(emake (pop argv))"
 
 ifndef ORG_LOAD_PATH
 ORG_LOAD_PATH  = -L ../../dash
--- a/emake.mk
+++ b/emake.mk
@@ -1,0 +1,35 @@
+-include config.mk
+include default.mk
+
+.PHONY: run clean install
+
+$(EMAKE_WORKDIR)/elpa: EMACS_ARGS += --batch
+$(EMAKE_WORKDIR)/elpa: $(EMAKE_WORKDIR)/emake.el
+	$(EMAKE) install
+
+$(EMAKE_WORKDIR)/emake.el:
+	@mkdir -p $(EMAKE_WORKDIR)
+	$(CURL) 'https://raw.githubusercontent.com/vermiculus/emake.el/$(EMAKE_SHA1)/emake.el' \
+	    --output '$(EMAKE_WORKDIR)/emake.el'
+
+clean:
+	@printf "Cleaning...\n"
+	@rm -rf $(EMAKE_WORKDIR)
+
+# Run Emacs with minimal configuration
+dev-run: $(EMAKE_WORKDIR)/emake.el $(EMAKE_WORKDIR)/elpa
+	@$(EMAKE_BASE) -f emake-setup-load-path
+
+# Like `dev-run', but always install from MELPA
+run: EMAKE_ENV += EMAKE_USE_LOCAL="NEVER"
+run: dev-run
+
+install: $(EMAKE_WORKDIR)/elpa
+
+compile: EMACS_ARGS += --batch
+compile: $(EMAKE_WORKDIR)/elpa $(EMAKE_WORKDIR)/emake.el
+	@$(EMAKE) compile
+
+loaddefs: EMACS_ARGS += --batch
+loaddefs: $(PACKAGE_LISP) $(EMAKE_WORKDIR)/emake.el
+	@$(EMAKE) autoloads
--- a/lisp/Makefile
+++ b/lisp/Makefile
@@ -1,14 +1,8 @@
 -include ../config.mk
 include ../default.mk
 
-lisp: $(ELCS) loaddefs
-
-loaddefs: $(PKG)-autoloads.el
-
-%.elc: %.el
-	@printf "Compiling $<\n"
-	@$(EMACS) -Q --batch $(EMACS_ARGS) \
-	$(LOAD_PATH) --funcall batch-byte-compile $<
+lisp: loaddefs $(ELS)
+	@cd .. && $(MAKE) -f emake.mk compile
 
 CLEAN  = $(ELCS) $(PKG)-autoloads.el
 
@@ -16,30 +10,6 @@ clean:
 	@printf "Cleaning...\n"
 	@rm -rf $(CLEAN)
 
-define LOADDEFS_TMPL
-;;; $(PKG)-autoloads.el --- automatically extracted autoloads
-;;
-;;; Code:
-(add-to-list 'load-path (directory-file-name \
-(or (file-name-directory #$$) (car load-path))))
-
-;; Local Variables:
-;; version-control: never
-;; no-byte-compile: t
-;; no-update-autoloads: t
-;; End:
-;;; $(PKG)-autoloads.el ends here
-endef
-export LOADDEFS_TMPL
-#'
-
+loaddefs: $(PKG)-autoloads.el
 $(PKG)-autoloads.el: $(ELS)
-	@printf "Generating $@\n"
-	@printf "%s" "$$LOADDEFS_TMPL" > $@
-	@$(EMACS) -Q --batch --eval "(progn\
-	(setq make-backup-files nil)\
-	(setq vc-handled-backends nil)\
-	(setq default-directory (file-truename default-directory))\
-	(setq generated-autoload-file (expand-file-name \"$@\"))\
-	(setq find-file-visit-truename t)\
-	(update-directory-autoloads default-directory))"
+	@cd .. && $(MAKE) -f emake.mk loaddefs


tarsius:
I've been thinking, maybe we should just put all the classes in a library that is loaded early on.

vermiculus:
What are the advantages? One disadvantages is the separation of a class's declaration from its implementation.

vermiculus:
Would an advantage be the removal of all --childp macros/functions? If so then that's probably worth it.

tarsius:
> Would an advantage be the removal of all --childp macros/functions? If so then that's probably worth it.

This functions would still get defined by `defclass`. Also, if we need a predicate function that takes one argument, then these functions are actually useful. So to me this isn't something that enters the consideration when thinking about collecting all classes in one place.

Also there sadly is this: https://github.com/magit/magit/issues/3732. By the way, if this issue starts to affect `forge` too and we cannot figure out under what circumstances `cl-typep` doesn't work as it supposed to and can convince it to give us the right answer, then I will likely bump the dependency to Emacs 26.1. For `transient` a little bit of uglyfication did the trick but for `forge` that wouldn't be feasible.

tarsius:
What we discussed in the past was using this for Magit but *only* for building on Travis.  I would still like to do that.

I haven't looked at Emake in quite a while and was planning to merge this to test it out while staying with my current approach for all my other packages.  But now that I have tried it out I have decided to stick to my approach.

I believe to remember that when I looked at Emake in more detail I liked it more than Cask.  So I would probably be satisfied if authors migrated from Cask to Emake.

However its main design decision seems to be the same as for Cask and I consider that to be a mistake.

Both utilities pull down the dependencies and put them somewhere inside the project whose dependencies they are.  That idea appears to be inspired by what is being done for other languages, for which I think doing it that way *does* make sense. But Emacs (Lisp) is fundamentally different and here, at least to me, doing it like this *only* has drawbacks.

With Ruby, Python, Javascript+Node etc. different applications are independent from each other.  Different applications can use different versions of their dependencies because they run in independent instances of the language runtime (or as separate system processes).

This is not the case with Emacs.  Of course you *can* start multiple instances, but the normal mode of operation is to have one instance and all "applications" run inside that one instance. It is not possible to use different versions of the same library for different "applications".

Therefore it does not make any sense IMO to make it possible to use different versions during development (and if it is only for building).  All this does is to hide potential conflicts that arise from using a newer/older version of some library than what will be used in the interactive Emacs instance.

----

Furthermore I try to break up my packages into small units to make them more usable by others.  That means that I often have to make changes across multiple repositories in parallel.  That becomes significantly harder if I have to build some wip-release of a modified library only so that I can test if it works together with another change in another repository.

----

I have a very strong opinion about this and it is unlikely to change.  I have been doing it my way for more than a decade and it has worked very well for me.  (And I realize the irony for criticizing Drew for doing the same.) I believe my approach is at the same time significantly simpler, powerful, and appropriate to the development of *Emacs* packages.

What is a best practice for other languages, which clearly separate different applications from each other, is not suitable for Emacs, which runs all applications in a single run-time and which has no support for modules or some other mechanism for versioning dependencies.

My approach only works well if all packages are in sub-directories of a common parent-directory.  (Though that isn't really a hard requirement, its just another instance of doing the easiest thing instead of paying the cost for making things more flexible.) It works best if those repositories are submodules.  While I think that every developer would benefit greatly by converting to my way of doing it, I realize that it would take a great amount of evangelizing to convince everyone to do so within the next decade or so.

But I am not willing to do that evangelizing.  I have however, over the course of a decade, written a tool that makes it easy to use this work-flow and a few years ago I released it.  The Borg package manager does not enjoy a great following, but those who stumbled across it and stuck with it long enough to understands what benefits it brings (by keeping things simple, but being highly opinionated) seem to enjoy it a great deal.

----

That being said it should be easy to contribute to my packages without following my approach.  In the case of Magit I have done that using the extremely ugly path-mangling kludges in its `default.mk`.

I plan to replace that fragile `make` boilerplate with some generalized lisp code.  I.e. I would like to replace

``` makefile

DASH_DIR ?= $(shell \
  find -L $(ELPA_DIR) -maxdepth 1 -regex '.*/dash-[.0-9]*' 2> /dev/null | \
  sort | tail -n 1)
ifeq "$(DASH_DIR)" ""
  DASH_DIR = $(TOP)../dash
endif

# ... many instances of the above ...

LOAD_PATH += -L $(DASH_DIR) -L $(OTHER_DIR) ...
```

with just

``` makefile
LOAD_PATH = $(EMACS) -Q --load get-load-path.el
```

vermiculus:
Parallel development of separate packages was definitely considered here. I should've provided an example of how one would approach this; I'll do that shortly. I would expect that to be done in config.mk, but it could easily be done in default.mk or elsewhere.

vermiculus:
I do agree with what you've said, though I also firmly believe it's important to occasionally test with how it will be installed in practice (which has traditionally been difficult, at least for me). So having the option of using only published versions of packages is important IMO.

I'm on my phone though and didn't want to disrespect such a thoughtful post with a short quip :-) just know that I appreciate the concern and believe there is already a solution.

tarsius:
> I also firmly believe it's important to occasionally test with how it will be installed in practice (which has traditionally been difficult, at least for me).

True and I have to admit I don't do that enough. I am starting to think that it might be a good idea to add an *additional* `emake` target, i.e. keep the current build system but add Emake as an alternative.

vermiculus:
My current plan is to modify the default behavior to look for packages in a parent directory first and only install them from an ELPA if they cannot be found there. A Make variable will be introduced to categorically ignore packages from other directories even if they exist. I'm going to try to do this all as a lisp file that's loaded with Emake. Would that meet the same goals as a separate `emake` target?

Though depending on what the goals are, a separate `emake` target may be all-in-all simpler – I'm just not certain it would work.

vermiculus:
I'm finding myself jumping through some hoops to support the lack of a `[Package-]Version` header – notably in closql. Upon calling `package-buffer-info` within `closql.el`, the function errors without this header.

Was the omission for Package-Version from closql's headers intentional? It's not terribly difficult to get around, but I thought I'd mention it.

tarsius:
> Was the omission for Package-Version from closql's headers intentional?

Yes because

1. If there is a Git repository then `git describe` gives a more correct response.
2. If installing using `package.el`, then Melpa took care of injecting the value.
3. Updating it would be annoying.
4. I would regularly forget to do it.

> Though depending on what the goals are, a separate `emake` target may be all-in-all simpler – I'm just not certain it would work.

Why would it not work? There are two, independent build targets -- everyone can choose the one that works for them. 

> My current plan is to modify the default behavior to look for packages in a parent directory first and only install them from an ELPA if they cannot be found there. A Make variable will be introduced to categorically ignore packages from other directories even if they exist. I'm going to try to do this all as a lisp file that's loaded with Emake.

A reasonable approach and I think you should experiment with it even though there is a risk that I won't merge it.

> Would that meet the same goals as a separate `emake` target?

Not really.

For example, a network connection is needed to download the Emake scripts. But if I were you I would implement it anyway (both approaches actually). If you want Emake to take off then you will have to take into account lots of diverging opinions on how things ought to be done. If nothing else, dealing with this particularly opinionated maintainer lets you dive right in. :grinning:

vermiculus:
> Package-Version header

I've made the necessary changes to not require Package-Version, though I do wish `package.el` made the patch instead.  I'm not entirely sure why it's necessary in practice.

> I would regularly forget to do it

I've had a very hard time with this too, as a matter of fact (vermiculus/magithub#330).  I'd love to hear how you avoid it :smile:

---

After a lot of noodling and a few tough problems, I've enhanced EMake
to respect a new environment variable, `EMAKE_LOCAL`.  Its default behavior is to install via `package.el` only those dependencies which cannot be found on disk by any function in a particular long-winded variable I've introduced (see `.build.el`).  This behavior can be overridden in the two obvious ways:
- A value of `ALWAYS` will refuse to install via `package.el` any dependency it cannot find locally.
- A value of `NEVER` will simply install everything via `package.el` as before.

I believe the new default behavior makes sense -- it minimizes network traffic when possible -- but these overrides allow for two important use-cases:
- `ALWAYS` tests everything as it would be seen if *everything else* were to be published
- `NEVER` tests everything as it would be seen if *nothing else* were to be published

tarsius:
That sounds useful but would be even more useful if there were a `run` command. If I understand the readme correctly, you don't want to add that?

vermiculus:
What would run do?

Despite what the README says, I've implemented something like `make emacs-Q` in the past. I'm not opposed to doing so these days as long as I control scope. I think the basics of how EMake works are stable enough to experiment with it again, though.

If I recall correctly, I believe I was able to do this without making lisp changes. I'll see if this is still feasible after the recent enhancements.

vermiculus:
I've squashed the commits and cleaned some things up as far as history. I've added a `run` command analgous to magit's `make emacs-Q` – I'm hoping that's what you meant.  Note the implementation of `run` (called `load` by EMake) is trivial:

```elisp
(defun emake-load ()
  "Load the testing environment.
This is intended to be run without the `--batch' invocation flag."
  (declare (emake-default-target "load"))
  (emake-with-elpa
   ;; `emake-with-elpa' ensures that `load-path' is setup correctly.
   (ignore)))
```

You could've added this as a custom target (as I've done with `make-autoloads` in `build.el`, but it's probably useful enough to include in the standard distribution.

vermiculus:
Rebased for gitignore conflicts.

tarsius:
> I've added a `run` command analgous to magit's `make emacs-Q` – I'm hoping that's what you meant.

That's what I had in mind.

----

The `.emake` directory is created at the top-level but the gitignore rule suggests you expect it to appear as `lisp/.emake`.

The `.emake` directory and a subdirectory per dependency is created even though I later get this for every dependency:

```
emake:INFO: Using dependency `X' at /home/jonas/.emacs.d/lib/X/
```

I think `.emake` should only be created if we are actually going to use it.

vermiculus:
Addressed the gitignore problem; neglected to update it after changes

You should only be getting one subdirectory of `.emake/elpa/` per dependency actually installed. See documentation of EMAKE_USE_LOCAL and its use in new file `emake.mk` for `make run` and `make dev-run` (which I could rename to `run-published` and `run`, respectively, at your preference – my shell has tab-completion so it's of little consequence to me).

tarsius:
Why is `build.el` needed? IMO other projects would need the same functionality too so this should be moved to Emake itself.

vermiculus:
I considered that too. Now that everything is merged into master, I can look at 'good' ways to approach this. I'm thinking of taking an approach similar to MELPA recipes, but in one data structure.

vermiculus:
On reading this again, do you think other projects need a way to generate autoloads as well? Is the handling you have significantly different than `package-generate-autoloads`?

tarsius:
> On reading this again, do you think other projects need a way to generate autoloads as well?

Absolutely!

> Is the handling you have significantly different than `package-generate-autoloads`?

No, and while it has been a while I think I actually copied it from there.

vermiculus:
Looks like it does indeed generate the same file with what I believe to be a syntactic difference:
```diff
diff -u /Users/sean/github/forge/autoloads-emake.el /Users/sean/github/forge/autoloads-custom.el
--- /Users/sean/github/forge/autoloads-emake.el	2019-02-26 19:57:01.000000000 -0600
+++ /Users/sean/github/forge/autoloads-custom.el	2019-02-26 19:57:18.000000000 -0600
@@ -1,10 +1,7 @@
 ;;; forge-autoloads.el --- automatically extracted autoloads
 ;;
 ;;; Code:
-
-(add-to-list 'load-path (directory-file-name
-                         (or (file-name-directory #$) (car load-path))))
-
+(add-to-list 'load-path "/Users/sean/github/forge/lisp")
 
 ;;;### (autoloads nil "forge-bitbucket" "forge-bitbucket.el" (0 0
 ;;;;;;  0 0))
@@ -288,6 +285,5 @@
 ;; version-control: never
 ;; no-byte-compile: t
 ;; no-update-autoloads: t
-;; coding: utf-8
 ;; End:
 ;;; forge-autoloads.el ends here

Diff finished.  Tue Feb 26 21:04:46 2019
```
So that's fun :-)

I've also introduced `emake-package-file-subdirectories` that can grow and be updated over time.

tarsius:
You cannot assume that the name of a repository of dependency matches the name of the package.

The package `emacsql-sqlite` for example does not have its own repository; it lives in the same repository as `emacsql`, and that is checked out at `../emacsql`. Same for `lv`, which is part of `hydra`.

tarsius:
I feel under pressure to merge this. You are not actively pressuring me but you have put quite some work into this and because of that I feel obliged to merge it eventually. Not least because I do not want to discourage you from contributing. 

So far this does not work well at all. I am only going to merge this if it works for me and you have to make sure it does. Please test this with the setup that I use and let me know once that actually works. The removed `DEPS` variable "documented" my setup.

To be honest, I am only considering to merge this to make it easier for you. I am only willing to do that if it doesn't make it much harder for me. Having to set `EMAKE_USE_LOCAL=ALWAYS` and much noisier output is acceptable.

Your approach does have advantages (including "test with what users will be using"). But it also greatly increases complexity. I can accept this complexity, but only if I don't have to deal with it myself. But Emake is not mature yet and quite a few unhandled edge-cases have come to light here.

I have said that I want to focus on helping others contributed to magit/forge/... for a while, but by that I did not mean that I want to help you develop Emake.

I mentioned that I want to implement something that works like `LOAD_PATH = $(EMACS) -Q --load get-load-path.el` eventually, but I don't want to do that now. When I work on my own stuff (as opposed to merge contributions), then I want to work on those things that I consider important right now). Implementing `get-load-path.el` is only important because that would allow you to easily compile and test your changes to Forge. I.e. I feel pressured to implement that now to ease some of the emake-related pressure.

----

I apologize for this outburst. I should have been more explicit about not being excited about this proposal much more explicitly.

At the same time you also have to realize that you did come in changing something very fundamental about the project. That's usually not a good thing to do when starting to contribute a project. I am fully aware that I have been guilty of doing this many times before myself--that doesn't change the fact that one shouldn't do it. And here we have a good example of why.

----

So what now?

I think you have to use the same setup as I use for a while. You might come to like it, but if that doesn't happen that's okay. But you have to do it because (1) you have to compile and test your other contributions and (2) you have to test that Emake works given this setup (because I am going to stick to it).

vermiculus:
Well, this is the second time I'm writing this post. I started on my phone, got three-ish paragraphs in, tapped the 'home' soft button (not an actual button), and it removed my post completely – no draft.

While I'll never remember what I said the first time, at least there won't be a third time.  This will not happen on Emacs :-)

---

First off -- you should feel under no pressure to merge this!  It will have no impact on my contributing to Forge in the future.  I'm certain I can find a way to use it without making changes to any other files -- that integration is all just Make in the end.  You shouldn't feel pressure to make things easier for me -- I am perfectly powerful to make them easier for myself.  If I do end up needing help, I'll let you know.

---

One of the things I tell mentees of mine is that just because something won't be used (e.g., a prototype of some new piece of the software) doesn't mean it's not worthwhile to do.  Software is a craft like any other -- while you might throw out a batch of cookies, it doesn't mean you didn't learn anything from the experience or that any part of it was necessarily wasted.  It's all about what your goals were going in.  If your goal was to give everyone cookies, then yeah, you failed; but if your goal was to learn about making cookies, you still succeeded.  I've learned a lot about `package.el` that I wouldn't have learned were it not for the requirements set forth by this particular maintainer :-)  That in itself is inherently valuable!

> A reasonable approach and I think you should experiment with it even though there is a risk that I won't merge it.

I took that at face value when you said it and I have no qualms about not merging this.  I'll go ahead and close this request as to clean up the list.  When it's good and ready (and after I've been testing it with different common development configurations -- including yours -- for longer), I'll re-open this pull request.  After all, I wouldn't want to waste such an even PR number as 100 :-)  This will probably be on the order of months from now.

---

I wouldn't necessarily consider the build tool to be fundamental to the project (I would consider that to be something like 'this database idea is all wrong'), but I do recognize it's a high-impact area.  I believe EMake allows for more flexibility and less duplication among projects and I will likely continue to submit pull requests implementing its use (only when replacing a much more complex/messy build system) as long as the relevant maintainers are amenable to the idea.

---

I wasn't aware that `lv` was part of `hydra`, though it makes obvious sense (now that you mention) it that some features are developed in shared repositories, so any build tool with similar goals as EMake should be able to handle that.  I can continue work to that point.  It's really only this idea of using local development copies that is unstable right now, so hopefully there is not too much work ahead.

vermiculus:
I have pushed the `emake` branch to my fork and am deleting the branch here for now.

tarsius:
Thanks for understanding. :kissing_heart:

vermiculus:
:heart:

PR #105, opened 2019-02-21
WIP Topic subsets

vermiculus:
Fix #79; Fix #88

I plan to implement this over the weekend; the current patches prepare for that.

One open question that I have is what you'd like to do for configuration.

> However instead of providing alternative section inserters I think I favor teaching the existing inserters to display different sets depending on settings that can set using configuration popups (similar to `L` and `D` for log and diff arguments) and to display the settings at the end of the list section heading.

I do plan on creating alternative section inserters in this pull request, but I also want to get more of your thoughts on your idea for a configurable catch-all section (which would remain the default).  Specifically, how do you imagine the configuration taking place?

- git variables (what I did in Magithub; wasn't too bad, but became difficult to manage)
- many `defcustom` variables (this could still grow to an untenable number of options with documentation strewn all about the place)
- one 'complex' `defcustom` variable per topic-type (issue, pullreq) – likely an alist of individual options
- something else?

(I'm looking forward to this opportunity to work with Transient :smile:)

PR #110, opened 2019-02-27
Optionally return rows of topics instead of objects

tarsius:
This is an alternative to #105.

`forge-topic-list-order` is currently ignored when returning rows.


tarsius:
@vermiculus I have merged the first three commits into `master`.  The remaining commit replaces #105.  I haven't merged that yet because there isn't any code yet that uses the new argument.  When you resume working on making topic listing more flexible, then you should base that work on this commit.

PR #113, opened 2019-03-04
Fix docstring

jabranham:
`forge-dispatch` is missing the arglist argument. This adds it, which lets forge-dispatch have documentation.


tarsius:
Thanks!

PR #122, opened 2019-03-15
Autoload forge-dispatch

jabranham:
This adds an autoload so that ' is bound to forge-dispatch in magit buffers without users explicitly needing to require forge.


tarsius:
Thanks!

I changed it depend on the feature instead of the filename and added two newline characters.

PR #127, opened 2019-03-22
forge-database-file: Use user-emacs-directory

Stebalien:
Otherwise, forge recreates ~/.emacs.d even if the user has moved their emacs directory elsewhere.

### DIFF
--- a/lisp/forge-db.el
+++ b/lisp/forge-db.el
@@ -29,7 +29,7 @@
 
 ;;; Options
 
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks for pointing this out, but that's not the correct way to do it either.

Stebalien:
You're right, that's definitely more correct (although, if this was working before, it shouldn't _really_ matter).

PR #128, opened 2019-03-24
Disable Markdown inline images in terminal

MatthewFluet:
Condition display of inline images on `(display-images-p)`

`(markdown-display-inline-images)` errors when the display cannot display images (e.g., terminal).  This causes `magit-refresh-buffer` to terminate early and subsequent navigation in a Topic buffer errors with `Error in post-command-hook (magit-section-update-highlight): (wrong-type-argument number-or-marker-p nil)`.


tarsius:
Thanks!

PR #136, opened 2019-04-13
Add Browse Commands to List Modes

PuercoPop:
I was surprised when M-x forge-browse-dwim didn't work when looking at the list of issues or pull-requests. This PR adds that functionality. Although I feel my current implementation is boiler-plate. Would have prefer to have written something like 

``` emacs-lisp
(defun forge-list-browse-topic ()
  (interactive)
  (forge-browse-topic (forge-get-topic (tabulated-list-get-id))))
```

And that that to the `FORGE-TOPIC-LIST-MODE-MAP`. Or `(browse-topic (forge-topic-at-point))`. Consider this PR a feature request with a little of extra effort.

P.D. Thanks for writing Forge.

### DIFF
--- a/lisp/forge-list.el
+++ b/lisp/forge-list.el
@@ -58,6 +58,7 @@
     (set-keymap-parent map forge-topic-list-mode-map)
     (define-key map (kbd "RET") 'forge-list-visit-issue)
     (define-key map [return]    'forge-list-visit-issue)
+    (define-key map (kbd "o")   'forge-list-browse-issue)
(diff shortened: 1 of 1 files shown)


tarsius:
That's good enough for now.

Sorry for the slow response; I completely forgot about this.

PuercoPop:
Don't worry about it. It wasn't perceived as slow on my end. Thanks for Forge, it has greatly improved my experience

PR #137, opened 2019-04-16
forges: add a new common gitlab instance "framagit"

paulRbr:
Framasoft is a well known non profit organisation (at least in France)
which offers a set of free (speech) software tools hosted for the well
being of public users. Details on https://framasoft.org/en/.

[Framagit](https://framagit.org/) is one of the service they offer to the public which is a
common Gitlab instance. (Similar to Debian's Salsa instance).

The intention of this PR is to add the framagit.org instance as a
forge.

I guess the way to add "gitlab.com-like forges" might need a special
feature inside Forge's codebase. There is a workaround to configure
local emacs configuration to use private instances seen in #9. But
maybe "well-known" and "important" public instances could still be
listed inside Forge's code.

WDYT? Thanks!

### DIFF
--- a/docs/forge.org
+++ b/docs/forge.org
@@ -108,6 +108,7 @@ also see [[*Getting Started]].
 
   - https://gitlab.com
   - https://salsa.debian.org
+  - https://framagit.org
 
 ** Partially Supported Forges
 :PROPERTIES:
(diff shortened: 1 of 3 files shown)


tarsius:
I welcome such additions.

Sorry for the slow response, this fell through the cracks somehow.

paulRbr:
@tarsius no problem. Thanks for merging this! 🙇‍♂️

PR #139, opened 2019-04-17
fix(labels): label containing UTF-8 characters made forge-pull crash

paulRbr:
The crash stacktrace looks like:

```
Debugger entered--Lisp error: (error "Multibyte character in data for base64 encoding")
  base64-encode-string("framagit.org:16:💪 Contribution Week" t)
  [...]
  forge--object-id("ZnJhbWFnaXQub3JnOjE2" "💪 Contribution Week")
  [...]
  forge--update-labels(#<forge-gitlab-repository forge-gitlab-repository> (((id . 38546) (name . "💪 Contribution Week") (color . "#69D100") (description . "Issues select...
```

I used the recommendation from this Emacs mailing list
https://lists.gnu.org/archive/html/help-gnu-emacs/2012-04/msg00173.html
to force the encoding with the `encode-coding-string` function before
calling the base64-encode function.

Testing on a fake personnal project
https://gitlab.com/paulrbr/test-markdown/issues/1 seems to work.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -130,22 +130,26 @@ This variable has to be customized before `forge' is loaded."
 (cl-defmethod forge--object-id ((class (subclass forge-topic)) repo number)
   "Return the id for a CLASS object in REPO identified by id NUMBER."
   (base64-encode-string
-   (format "%s:%s%s"
-           (base64-decode-string (oref repo id))
-           (substring (symbol-name class)
-                      (length (oref-default class closql-class-prefix)))
-           number)
+   (encode-coding-string
+    (format "%s:%s%s"
+            (base64-decode-string (oref repo id))
+            (substring (symbol-name class)
+                       (length (oref-default class closql-class-prefix)))
+            number)
+    'utf-8)
    t))
 
 (cl-defmethod forge--object-id ((prefix string) id)
   (base64-encode-string
-   (format "%s:%s"
-           (base64-decode-string prefix)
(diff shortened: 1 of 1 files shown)


tarsius:
Please put `(encode-coding-string` on its own line.

paulRbr:
> Please put `(encode-coding-string` on its own line.

Indeed, readability is enhanced :). I have fixed your recommendation.

Thanks for the review!

tarsius:
Something doesn't seem right.  I'll have to look into this in more detail.

paulRbr:
> Something doesn't seem right. I'll have to look into this in more detail.

I am new to forge's codebase so it would completely make sense if it doesn't seem right. This bug fix was done with a naïve approach until I get it working for my use-case. 

Please let me know if you want me to look at something special or if I could help for anything.

tarsius:
I think the string that is being encoded should always be of a "number", so it shouldn't contain any characters that need encoding. Maybe that isn't the case for gitlab or maybe forge somewhere stores the wrong value in the `number` slot (sounds like it could be the title).

noname007:
i meet this problem when gitlab labes  is chinese

noname007:
thanks for your code 😀 @paulRbr

tarsius:
... trying out something unrelated ...

tarsius:
I have finally applied this--sorry for taking so long.

I have kept your commit (with attribution) but have written my own commit message, which explain why this is needed.  The TL;DR is that we only need this because of a defect of the Gitlab API concerning labels.

PR #145, opened 2019-05-08
Fix 'missing' function forge-list-browse-pullreq

PuercoPop:
This is fix for #136. The name `forge-list-browse-issue` is used for both
commands resulting in the function `forge-list-browse-pullreq` to be
missing. Sorry for the oversight.

### DIFF
--- a/lisp/forge-list.el
+++ b/lisp/forge-list.el
@@ -119,8 +119,8 @@
   (interactive)
   (forge-visit-pullreq (forge-get-pullreq (tabulated-list-get-id))))
 
-(defun forge-list-browse-issue ()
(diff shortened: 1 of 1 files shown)


lisp/forge-list.el
```diff
@@ -119,7 +119,7 @@
   (interactive)
   (forge-visit-pullreq (forge-get-pullreq (tabulated-list-get-id))))
 
-(defun forge-list-browse-issue ()
+(defun forge-list-browse-pullreq ()
   "Visit the url corresponding to the issue at point in a browser."
```
dakra: Should be pull request in the docstring as well?!

tarsius:
Thanks to both of you.

PR #151, opened 2019-05-19
Add upper case to github PR template files

pkryger:
`PULL_REQUEST_TEMPLATE` is also a valid file name in addition to
`pull_request_template`. Frankly, I've never encountered the lower
case version until I was poking around why `forge` is not picking up
my project's templates.

### DIFF
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -471,7 +471,8 @@ repositories.
                                       (_ (subclass forge-issue)))
   (when-let ((files (magit-revision-files (oref repo default-branch))))
     (if-let ((file (--first (string-match-p "\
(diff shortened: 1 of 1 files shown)


vermiculus:
Wouldn't a corresponding change need to be made for issues?

All-in-all, it may be simpler to manipulate the casing of the files themselves rather than to extend the regex language.

pkryger:
@vermiculus that's a very good point. However, I'm not really familiar with the API nor the conventions here. Do you know if both of them should be changed (i.e., the [file](../blob/0375541/lisp/forge-github.el#L474) and [directory](../blob/0375541/lisp/forge-github.el#L477))?

vermiculus:
I'm not sure. I would refer to [the docs](https://help.github.com/en/articles/about-issue-and-pull-request-templates).

tarsius:
I have addressed this by binding `case-fold-search` to `t` (which is the default but can be changed by the user).

PR #167, opened 2019-06-29
Fix some doc typos

stepnem:
I also suggest you update the HTML version at
https://magit.vc/manual/forge/, I was tripped up by the `F y` error
(fixed in c0a5a449 2 months ago) still in there.


tarsius:
Thanks!

PR #168, opened 2019-07-06
Display base and head branch refs for pull requests

darth10:
This is an attempt to show branch refs for PRs.
I find this information useful as I often need to confirm which branch the PR is made from and to.

`Refs` is probably not the best name for this, but that's what it's defined as by GH API.
More than happy to change this. 

_Sample output:_

```
Refs:      sigma/gh.el:master...yuya373/gh.el:set-async-from-api
```

image


darth10:
Hey @tarsius, 
I was hoping this could be merged in soon.
If there are any issues with this change, please let me know.

PR #196, opened 2019-09-16
Fix forge-delete-post

### LINKED ISSUE #195: Delete post does not work - and is called "delete comment"
Comments on issues and PRs are calles "posts" everywhere(?) except when deleting them, then they are called "comments":  See forge.org section "Editing Topics and Posts".  The command is also called `forge-delete-comment`.  The `forge-delete-comment` command uses `forge-comment-at-point` instead of `forge-post-at-point`. `forge-delete-comment` is the only user of `forge-comment-at-point`, which is a good thing as that function is broken. It returns `t` instead of the `forge-post` object at point.

I think I got all of that right :)

snogge:
`forge-delete-post' used to be `forge-delete-comment', and it was
broken.

* Rename `forge-delete-comment' -> `forge-delete-post'
* Rename `forge--delete-comment' -> `forge--delete-post'
* Move common code from `forge--delete-post' to `forge-delete-post'
* Remove `forge-comment-at-point', no longer used

Closes #195.

#

### DIFF
--- a/docs/forge.org
+++ b/docs/forge.org
@@ -728,9 +728,9 @@ a post visits that post in a browser.
   This command reads a list of people who you would like to review an
   existing topic in the minibuffer.
 
-- Key: C-c C-k [on a comment section], forge-delete-comment
+- Key: C-c C-k [on a post section], forge-delete-post
 
-  This command deletes the comment at point.
+  This command deletes the post at point.
 
 Creating a new post and editing an existing post are similar to
 creating a new commit and editing the message of an existing commit.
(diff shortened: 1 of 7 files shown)


tarsius:
> Move common code from `forge--delete-post` to `forge-delete-post`

That's not correct. `forge--delete-post` could be called elsewhere and has to cover the complete task of deleting the post. The `magit-refresh` should probably be moved, but I'll leave it where it is for the time being (because other low-level functions currently also refresh and I want to fix that in one go).

PR #199, opened 2019-09-28
Avoid breaking magit when sqlite binary is not available (fix #84)

### LINKED ISSUE #84: Using forge in Windows breaks magit
Here is my configuration:
https://github.com/drrlvn/.emacs.d/blob/master/init.el#L792

It works well in macOS and Linux. 

When running it in Windows, whenever I call `magit-status` an empty buffer opens and `*Messages*` shows:

```
Could not find C compiler, skipping SQLite build
emacsql-sqlite-ensure-binary: No EmacSQL SQLite binary available, aborting
```

It seems to be explained [here](https://github.com/skeeto/emacsql#windows-issues). I'd understand if supporting Windows is too much of an effort, but perhaps it should be documented in the README if so.

r-darwish:


tarsius:
* The name of the variable doesn't seem right.
* Use `with-demoted-errors` instead of `ignore-errors`.

tarsius:
I didn't mean the typo--I didn't even notice that.
The name should end with `-p`.
But what I was really thinking about is that this isn't about whether "the db" is available or not, but whether the sqlite binary is, so I think `forge--sqlite-available-p` should do. If you have a better suggestion though...

Also please put the `with-demoted-errors` on a new line. I like the look of that better.

tarsius:
Cross reference #84 and #539.

PR #210, opened 2019-10-26
forge-get-repository: Add missing arg in generic.

cireu:
----

#

### DIFF
--- a/lisp/forge-core.el
+++ b/lisp/forge-core.el
@@ -119,7 +119,7 @@ If this is nil, then `forge-pull-notifications' has to be used."
 The hierarchy is repository > topic > post.
 For other objects return nil.")
 
-(cl-defgeneric forge-get-repository ()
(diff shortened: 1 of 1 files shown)


tarsius:
There isn't actually a single method with exactly this signature and only one method takes DEMAND as the first argument. The reason this argument is documented here is that several methods take such an argument *somewhere* and documenting it here means we only have to do it once.

In light of that I don't think it not appropriate to add DEMAND to the signature of the generic function (which by the way only serves as documentation anyway).

cireu:
I see, the arg of `forge-get-repository` have both internal and external usage and each usage have variant meanings, right?

But I still suggest you add an placeholder in arglist of generic to mind people send an arg to call this function.

tarsius:
I'll think about it.

tarsius:
I ended up merging this after all.

Issue #241, opened 2020-01-29
Only a single PR review-request is listed for each PR


huonw:
Version: `20200126.1315`

I'm experimenting with a way to view the PRs which are awaiting a review from me, by adjusting the `forge-*-assigned-pullreqs` functions:

```elisp
(defun forge-insert-review-requested-pullreqs ()
  "Insert a list of open pull-requests that are awaiting review by you."
  (when-let ((repo (forge-get-repository nil)))
    (unless (oref repo sparse-p)
      (forge-insert-topics "Pull requests awaiting review"
                           (forge--ls-review-requested-pullreqs repo)
                           (forge--topic-type-prefix repo 'pullreq)))))

(defun forge--ls-review-requested-pullreqs (repo)
  (mapcar (lambda (row)
            (closql--remake-instance 'forge-pullreq (forge-db) row))
          (forge-sql
           [:select $i1 :from pullreq
                    :join pullreq_review_request :on (= pullreq_review_request:pullreq pullreq:id)
                    :join assignee         :on (= pullreq_review_request:id      assignee:id)
                    :where (and (= pullreq:repository $s2)
                                (= assignee:login     $s3)
                                (isnull pullreq:closed))
                    :order-by [(desc updated)]]
           (vconcat (closql--table-columns (forge-db) 'pullreq t))
           (oref repo id)
           (ghub--username repo))))
```

But this only lists a subset of the PRs that are awaiting a review from me (according to https://github.com/pulls/review-requested).

It looks like this is because the database only includes a single reviewer for each PR: the following query gets the number of review requests for each PR, and takes the maximum across all PRs; I ran it against `~/.emacs.d/forge-database.sqlite`:

```sql
sqlite> SELECT max(count) FROM (SELECT count(*) AS count FROM pullreq_review_request GROUP BY pullreq);
1
```

However, all of the repos connected to my database have PRs with multiple reviewers, e.g. https://github.com/stellargraph/stellargraph/pull/738 currently has 3 reviewers, with 2 outstanding, that hadn't reviewed at all (at the time of running the query above). To verify, this lists the reviewers of that PR, :

```sql
 SELECT login FROM pullreq INNER JOIN pullreq_review_request ON pullreq.id = pullreq_review_request.pullreq INNER JOIN assignee ON assignee.id = pullreq_review_request.id WHERE number = 738;
"kjun9"
```

tarsius:
[...]

Issue #82, opened 2019-01-19
Support for "reviewers" as well as "assignees"?


stig:
Could we please have support for setting "reviewers" as well as assignees? The company I work for do not use assignees at all (it's implicit that the PR creator owns it, and is thus assigned) but we use reviewers extensively to request reviews from specific people.

tarsius:
So I thought "lets quickly tackle this", expecting to be able to pretty much copy-paste the assignee implementation. It seems I was wrong. "Reviewers" appear to be more complex and less intuitive than assignees, and also not sufficiently documented.

I'll need to reverse engineer this. Could you please point me to a few fairly active but preferably small repositories making extensive use of review requests?

glittershark:
@tarsius Do you mean via Github's API? The following GQL query works for me:
```
repository(owner: "magit", name: "forge") {
    pullRequests(first:5) {
      nodes {
        reviewRequests(first:5) {
          nodes {
            requestedReviewer {
              ...on User {
                name
              }
            }
          }
        }
      }
    }
  }
```

tarsius:
No I mean a *repository* that actually uses review requests so that I can test whether the code that I have written actually works. (And yes, I have figured out too that I have to use inline fragments (and implement support for that in `gsexp.el`.)

tarsius:
Question that I have are:

- What happens if the requested review is performed?
- Does the respective `requestedReviewer` disappear?
- If not, then why is there no `status` field on `requestedReviewer`?
- Does it matter if the pr is approved or "needs work IMO"?

stig:
Here is a few modestly-sized public repos that uses review requests: 

- https://github.com/CircleCI-Public/slack-orb
- https://github.com/CircleCI-Public/circleci-cli
- https://github.com/CircleCI-Public/circleci-orbs

(I've no affiliation -- though I have contributed to at least one.)

tarsius:
I have added basic support for *review requests*. Only Github is supported because Gitlab doesn't have that feature. You can request a review, see who was asked to perform a review while looking at it in a dedicated buffer, and revoke a request. Basically you can do what you can do when assigning an issue except that the feature is called "review requests" instead of "assignees". In other words what has been requested here.

There is no support for actually performing a review (~ #75) or listing prs you have been asked to review (~ #79). At least the latter I plan to tackle soonish.

dcaixinha:
[...]

PR #244, opened 2020-02-17
support multiple requested reviewers, display PRs waiting for you

achempion:
references #241, #79, #82

### DIFF
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -225,8 +225,8 @@
         (when bump
           (forge--set-id-slot repo pullreq 'assignees .assignees)
           (forge--set-id-slot repo pullreq 'review-requests
-                              (--map (cdr (cadr it))
-                                     (car .reviewRequests)))
+                              (--map (cdr (cadr (car it)))
+                                     .reviewRequests))
           (forge--set-id-slot repo pullreq 'labels .labels))
         pullreq))))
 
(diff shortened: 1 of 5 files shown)


lisp/forge-github.el
```diff
@@ -225,8 +225,8 @@
         (when bump
           (forge--set-id-slot repo pullreq 'assignees .assignees)
           (forge--set-id-slot repo pullreq 'review-requests
-                              (--map (cdr (cadr it))
-                                     (car .reviewRequests)))
+                              (--map (cdr (cadr (car it)))
+                                     .reviewRequests))
```
tarsius: When I saw that `car` I also thought "ah!! easy fix", but ... *now I got it*:

https://github.com/stellargraph/stellargraph/pull/738 actually only has one review request. One *outstanding* review requests and two (already performed) *reviews*.

Okay then! In that case please fix this bug in one commit and use another commit to add the new feature below.

tarsius: Oh, and please point me to a pull-request that currently has two open review requests so that I can verify myself that this fix really works.

achempion: https://github.com/kubernetes/kubernetes/pull/88249

tarsius:
Anyway, I put the suffix "on level 7" and added it to the manual.

achempion:
> Is there a way to "approve" a pull-request after it has already been merged?

I think you can just send a review but without additional options as approve / request changes.

Anyway, thank you for the library. 🚀

PR #245, opened 2020-02-24
forge--pullreq-branch handle: handle deleted github PR branch

MartinNowak:
- fixes type error triggered by `(string-match-p ":" nil)`


tarsius:
Thanks!

Issue #109, opened 2019-02-26
Forge Suffixes Don't Show Up In Magit Pop-ups


parhamdoustdar:
I'm using the latest Melpa version of Forge, and I'm not doing anything extra about it in my init file (i.e. not using `require` explicitly because installed packages are automatically required).

However, transient suffixes are not showing up. For example, when I visit the branching buffer by pressing `b`, I don't see a `y` key for `forge-checkout-pullreq`. Maybe the autoload cookie is missing and that line (which i found in `forge.el`) is never being executed?

I've checked the info pages, but I'm not seeing anything about extra configuration steps. Am I missing something here?

Thanks!

tarsius:
Yesterday I added this to the manual.  (Online version is awaiting cache invalidation.)

Loading Magit doesn't cause Forge to be loaded automatically.  Adding something like this to your init file takes care of this:

```emacs-lisp
(with-eval-after-load 'magit
  (require 'forge))

```

Or if you use ~use-package~:

```emacs-lisp
(use-package forge
  :after magit)

```

tarsius:
/cc @kyleam @purcell

This keeps coming up and I am wondering whether we should just autoload the first form, similar to what Kyle has done for his extensions to Magit. Steve sometimes argues against packages imposing themselves just because they happen to be installed, but I feel this might be a different situation. Simply loading `forge` doesn't change anything (*) except adding key bindings, and `magit` still has to be loaded by the user; either explicitly using `require`, which is not recommended, or by invoking an autoloaded command such as `magit-status`.

(*) For the time being at least, but that might change. Some of Magit's fetch commands might be taught to *also* fetch the forge metadata (but would only do so for package whose forge metadata was previously fetched explicitly).

parhamdoustdar:
I would personally support this. When I install a package, it would be strange for me to go somewhere else and say that I *really* want it to be loaded. If I don't want it, there are a lot of other places that I could modify to have it not be loaded.

vermiculus:
I disagree. I can see a case for not wanting forge features in magit but wanting access to its codebase for other development. How would you avoid installing features if they're autoloaded?

parhamdoustdar:
Good point. I know that `org` solves this by having a customizable list of
modules that it would load. Do you guys think it's useful to have something
like this in magit, where there is one place to register plugins that would
want to register their own prefix keys?
I mean, in the context of magit, it makes sense to have `magit-status`
autoloaded, because it's pretty much the gateway to magit functionality.
[...]

PR #248, opened 2020-02-27
Add gitlab.com forking support

stepnem:
(Depends on https://github.com/magit/ghub/pull/109)

### DIFF
--- a/lisp/forge-gitlab.el
+++ b/lisp/forge-gitlab.el
@@ -542,6 +542,15 @@ it is all or nothing.")
   (--filter (string-match-p "\\`\\.gitlab/merge_request_templates/.+\\.md\\'" it)
             (magit-revision-files (oref repo default-branch))))
 
+(cl-defmethod forge--fork-repository ((repo forge-gitlab-repository) fork)
+  (with-slots (owner name) repo
+    (forge--glab-post repo (format "/projects/%s%%2F%s/fork" owner name)
+                      (and (not (equal fork (ghub--username (ghub--host nil))))
+                           `((namespace . ,fork)))
+                      :noerror t)
+    (ghub-wait (format "/projects/%s%%2F%s" fork name)
+               nil :auth 'forge :forge 'gitlab)))
+
 ;;; Utilities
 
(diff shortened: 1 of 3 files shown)


lisp/forge-gitlab.el
```diff
@@ -542,6 +542,16 @@ it is all or nothing.")
   (--filter (string-match-p "\\`\\.gitlab/merge_request_templates/.+\\.md\\'" it)
             (magit-revision-files (oref repo default-branch))))
 
+(cl-defmethod forge--fork-repository ((repo forge-gitlab-repository) fork)
+  (with-slots (owner name) repo
+    (let ((resource (format "/projects/%s%%2F%s" fork name)))
+      (if (glab-get resource)
+          (forge--msg nil t nil "Repo %s/%s already exists!" fork name)
+        (forge--glab-post repo (format "/projects/%s%%2F%s/fork" owner name)
```
tarsius: The github variant intentionally does not check whether the fork already exists. That way `forge-fork` can be (ab)used to add a remote for a certain fork regardless of whether that fork already exists (and the user having to remember whether that is so or not).

stepnem:
>> +      (if (glab-get resource)
> +          (forge--msg nil t nil "Repo %s/%s already exists!" fork name)
> +        (forge--glab-post repo (format "/projects/%s%%2F%s/fork" owner name)
>
> The github variant intentionally does not check whether the fork
> already exists. That way `forge-fork` can be (ab)used to add a remote
> for a certain fork regardless of whether that fork already exists (and
> the user having to remember whether that is so or not).

Yes, but precisely in order to _preserve_ that forge-fork semantics, we
do have to check this for GitLab, because she returns 4XX error (409
Conflict IIRC) if you try to create an already existing project. Letting
the error through would prevent the subsequent remote manipulation step.

tarsius:
I overlooked that this is just a warning, not an error.

I think it would be better to avoid the additional request and just use the `:noerror` argument instead.

stepnem:
> I overlooked that this is just a warning, not an error.
>
> I think it would be better to avoid the additional request and just use the `:noerror` argument instead.

Indeed, I didn't realize I could use :noerror, thanks.

tarsius:
Thanks!

PR #249, opened 2020-02-28
Favour default-branch over master when creating a PR

pkryger:
Some repositories are not using `master` as their default branch. When there's
no upstream branch, I think it may be a good idea to fall back to the
`default-branch` for the repository before `master`.

I've also considered alternatives with [symbolic-ref of `HEAD`](https://stackoverflow.com/a/44750379/519827) and  [API]( https://stackoverflow.com/a/16501903/519827), yet given the simplicity of the solution in the b3de608 I decided to give it a shot. Please let me know what you think.


tarsius:
Thanks!

PR #272, opened 2020-05-05
Convert (oset obj key val) to (setf (oref obj key) val)

leungbk:
oset is obselete as of Emacs 28.1.

See https://github.com/magit/magit/issues/4105.

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -97,7 +97,7 @@ If pulling is too slow, then also consider setting the Git variable
            (format "Always pull all of %s/%s's topics going forward?"
                    (oref repo owner)
                    (oref repo name)))
-          (oset repo selective-p nil)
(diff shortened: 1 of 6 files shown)


leungbk:
I realized that `closql` adds advice to `eieio-oset`, so I'll close this PR under the assumption that it will introduce bugs.

basil-conto:
> I realized that `closql` adds advice to `eieio-oset`, so I'll close this PR under the assumption that it will introduce bugs.

Calling `(oset ...)` is the same as calling `(eieio-oset ...)` with a quoted `slot`, and is also what `(setf (oref ...) ...)` expands to.

tarsius:
[I am trying to get it un-deprecated.](https://lists.gnu.org/archive/html/emacs-devel/2020-05/msg00674.html)

PR #291, opened 2020-08-09
forge-pkg.el: Add homepage URL

winny-:
When browsing MELPA via `list-packages`, there is no homepage URL for
forge. I *think* this is how one adds an URL to the `define-package`
form, as suggested by this
[GitHub search](https://github.com/search?l=Emacs+Lisp&q=define-package+https%3A%2F%2F&type=Code).

### DIFF
--- a/forge-pkg.el
+++ b/forge-pkg.el
@@ -13,4 +13,5 @@
     ;; also remove the related kludge from Forge's Melpa recipe.
     (magit          "20190408")
     (markdown-mode  "2.3")
-    (transient      "0.1.0")))
+    (transient      "0.1.0"))
+  :url "https://github.com/magit/forge")


tarsius:
Not sure that's useful in this case; the <kbd>Source code</kbd> button already takes the user to that page. IMO <kbd>Homepage</kbd> should take the user to some other, less technical page. Ideally the homepage, but if there is none, as in this case, then the manual would do.

winny-:
I don't seem to have those buttons available in my Emacs package UI. I am using GNU Emacs 26.3. With the current defaults, it appears one has to locate forge's website from content available outside of emacs (such as melpa.org's webapp).

----

This is the information available from my emacs with an empty HOME and the following elisp evaluated to set up MELPA:

```
(require 'package)
(add-to-list 'package-archives
             '("melpa" . "http://melpa.org/packages/") t)
(when (< emacs-major-version 24)
  ;; For important compatibility libraries like cl-lib
  (add-to-list 'package-archives '("gnu" . "http://elpa.gnu.org/packages/")))
(package-initialize)
```

info

Here's another package with a Homepage URL:

info2

tarsius:
Didn't think of that. I don't actually use package.el and my own package manager [`borg`](https://github.com/emacsorphanage/borg) is smart enough not to require any hand holding here. :yum: 

Before I merge could you please remove the speculation at the end of the commit message, I am fairly sure it works that way. Just remove "I *think* ..."

winny-:
Great! Done. I  honestly have no clue what I'm doing :wink:, so I am pleased this PR appears to address the issue.

*Takes a peek at borg & epkg.* I hope this gets broader adoption, it seems a bit easier to work with than the defaults.

PR #296, opened 2020-09-26
Allow toggling Forge sections

dcluna:
Depending on the size of the notifications for the repo, the redisplay of the magit-status buffer is noticeably slower (in my particular case, the `magit-insert` hooks for PRs and issues regularly spend more than 2s each).

This PR allows the user to toggle redisplay for these sections as needed, to avoid this performance hit.

(As an aside - thanks for all your work on Magit & Emacs in general. Big fan.)

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -67,7 +67,9 @@ for a repository using the command `forge-add-pullreq-refspec'."
     ("c f" "fork or remote" forge-fork)]]
   [["Configure"
     ("a" "add repository to database" forge-add-repository)
-    ("r" "forge.repository" forge-forge.remote)]])
+    ("r" "forge.repository" forge-forge.remote)
+    ("t" forge-forge.topics)]])
+
 
 ;;; Pull
 
@@ -803,6 +805,19 @@ is added anyway.  Currently this only supports Github and Gitlab."
   :variable "forge.remote"
   :choices 'magit-list-remotes
   :default "origin")
+
+(defvar-local forge-display-topics t
+  "Whether the Forge topics are inserted in the Magit status buffer upon refresh.")
+
+(transient-define-suffix forge-forge.topics ()
+  :transient nil
+  :description (lambda ()
+                 (if forge-display-topics
+                     "hide topics"
+                   "display topics"))
+  "Hides/displays Forge topics in the Magit status buffer upon refresh."
+  (interactive)
+  (setq forge-display-topics (not forge-display-topics)))
 
 ;;;###autoload
 (defun forge-list-notifications ()
--- a/lisp/forge-pullreq.el
+++ b/lisp/forge-pullreq.el
@@ -246,7 +246,7 @@ yourself, in which case you probably should not reset either.
   "Insert a list of mostly recent and/or open pull-requests.
 Also see option `forge-topic-list-limit'."
   (when-let ((repo (forge-get-repository nil)))
-    (unless (oref repo sparse-p)
+    (when (and forge-display-topics (not (oref repo sparse-p)))
       (forge-insert-topics "Pull requests"
                            (forge-ls-recent-topics repo 'pullreq)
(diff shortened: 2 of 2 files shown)


tarsius:
I think this should be a single command, it shouldn't talk about "toggling" (because in this context that already means "collapse or expand"), and it should not manipulate the hook (because we don't know for sure where to reinsert) instead it should change a boolean variable which is respected by the section inserters (there are other inserters that already do this).

lisp/forge-pullreq.el
```diff
@@ -246,7 +246,7 @@ yourself, in which case you probably should not reset either.
   "Insert a list of mostly recent and/or open pull-requests.
 Also see option `forge-topic-list-limit'."
   (when-let ((repo (forge-get-repository nil)))
-    (unless (oref repo sparse-p)
+    (if (and forge-display-sections (not (oref repo sparse-p)))
```
tarsius, line 249: When there is no ELSE part, then use `when`.

lisp/forge-commands.el
```diff
@@ -67,7 +67,12 @@ for a repository using the command `forge-add-pullreq-refspec'."
     ("c f" "fork or remote" forge-fork)]]
   [["Configure"
     ("a" "add repository to database" forge-add-repository)
-    ("r" "forge.repository" forge-forge.remote)]])
+    ("r" "forge.repository" forge-forge.remote)
+    ("s" (lambda ()
```
tarsius, line 71: I would like to avoid a lambda here. If you define `forge-toggle-display-sections` using `define-transient-suffix`, then you could move this here. You could also define this as a named function, but that might be a bit much.

lisp/forge-commands.el
```diff
@@ -67,7 +67,12 @@ for a repository using the command `forge-add-pullreq-refspec'."
...
+           (if forge-display-sections
+               "hide PRs/issues"
```
tarsius, line 73: Just call them "topics".

lisp/forge-commands.el
```diff
@@ -67,7 +67,12 @@ for a repository using the command `forge-add-pullreq-refspec'."
...
+             "display PRs/issues")) forge-toggle-display-sections)]])
+
```
tarsius, line 75: Generally avoid following a multi-line s-expression by another s-expression without a newline character in between.

lisp/forge-commands.el
```diff
@@ -931,6 +936,14 @@ heavy development."
     (emacsql-enable-debugging db)
     (switch-to-buffer-other-window (emacsql-log-buffer db))))
 
+(defvar-local forge-display-sections t
+  "Whether the Forge sections are inserted in the Magit status buffer upon refresh.")
+
+;;;###autoload
+(defun forge-toggle-display-sections ()
+  (interactive)
+  (setq forge-display-sections (if forge-display-sections nil t)))
```
tarsius, line 945: This should be documented and maybe you can think of a better name.

tarsius (changes requested):

dcluna:
Took another stab at it, the code seems simpler after I implemented your suggestions. Can you take a look when possible?

Thanks,

lisp/forge-commands.el
tarsius: You added a doc-string to the variable, but what really needs one is this command. :grinning:

tarsius: Oh and to flip a boolean you can just use `not` around the current value.

dcluna: Hope I got the right doc-string this time :)

tarsius:
This still needed quite a few improvement and in the end I decided to attribute the commit to myself.

tarsius:
Note to past/future self: I am painfully aware that this is to sort of performance "fix" that was used by the old "list branches in a buffer" feature that I so loudly complained about when I finally removed it.

PR #302, opened 2020-11-03
Improve setup docs

zainab-ali:
----

This does a few things:
 - Adds an example for setting the `forge.remote` git config variable
 - Removes the reference to the wizard in the `Initial Pull` section.
 
   Previously, this read:
 
   ```
   When adding the first repository from https://github.com to your local
   database you will be guided through the process of creating the API token
   ```
 - Swaps around the `Initial Pull` and `Token Creation` sections.  There isn't a wizard anymore, so the token must be created first.

#

### DIFF
--- a/docs/forge.org
+++ b/docs/forge.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Forge: (forge).
 #+TEXINFO_DIR_DESC: Access Git Forges from Magit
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

zainab-ali:
Thanks!  Apologies for missing the texinfo file.  If I make any more doc PRs, I'll be sure to look at that first.

tarsius:
It's okay if you skip that. If I need to change something, then it actually makes it easier for me if you didn't re-generate the texi files.

See https://github.com/magit/magit/wiki/Documentation-tools-and-conventions.

PR #317, opened 2020-12-08
Add browse file support

raszi:
This pull-request adds a new feature to forge to be able to open in the browser / kill the URL of the actual file on the specific location for all the supported Git providers.

Adds the following functions:

```elisp
(forge-buffer-location)
```

```elisp
(forge-buffer-location-as-kill)
```

```elisp
(forge-browse-buffer)
```

### DIFF
--- a/lisp/forge-repo.el
+++ b/lisp/forge-repo.el
@@ -38,6 +38,7 @@
    (pullreq-post-url-format   :initform nil :allocation :class)
    (commit-url-format         :initform nil :allocation :class)
    (branch-url-format         :initform nil :allocation :class)
+   (file-url-format           :initform nil :allocation :class)
    (remote-url-format         :initform nil :allocation :class)
    (create-issue-url-format   :initform nil :allocation :class)
    (create-pullreq-url-format :initform nil :allocation :class)
--- a/lisp/forge-bitbucket.el
+++ b/lisp/forge-bitbucket.el
@@ -35,10 +35,18 @@
    (pullreq-post-url-format   :initform "https://%h/%o/%n/pull-requests/%i#comment-%I")
    (commit-url-format         :initform "https://%h/%o/%n/commits/%r")
    (branch-url-format         :initform "https://%h/%o/%n/branch/%r")
+   (file-url-format           :initform "https://%h/%o/%n/src/%r/%f#lines-%l")
    (remote-url-format         :initform "https://%h/%o/%n/src")
    (create-issue-url-format   :initform "https://%h/%o/%n/issues/new")
    (create-pullreq-url-format :initform "https://%h/%o/%n/pull-requests/new")))
 
 ;;; _
+
+(cl-defmethod forge--file-url ((repo forge-bitbucket-repository) rev file start end)
+  (let* ((start-fragment (number-to-string start))
+         (location-fragment (if end (concat start-fragment ":" (number-to-string end)) start-fragment)))
+    (forge--format repo 'file-url-format
+                   `((?r . ,rev) (?f . ,file) (?l . ,location-fragment)))))
+
 (provide 'forge-bitbucket)
 ;;; forge-bitbucket.el ends here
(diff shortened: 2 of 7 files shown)


lisp/forge-repo.el
```diff
@@ -351,7 +352,7 @@ Return the repository identified by HOST, OWNER and NAME."
     ;; of https.
     ((forge-gitlab-repository forge-gitlab-http-repository) 'gitlab)
     (forge-github-repository    'github)
-    (forge-gitea-repository     'gittea)
+    (forge-gitea-repository     'gitea)
```
raszi, line 355: This was a sneaky bug in the codebase.

tarsius:
> Merge branch 'master' into browse-file

Please rebase instead of merging. I have just done this now and then pushed the result to your branch.

tarsius:
While working on a feature branch, it is best to not "preserve mistakes". E.g. it is completely irrelevant that you made a typo in the first commit and then fixed it like so in the second commit:

```diff
-   (file-url-format           :initform "https://%h/%o/%n/-/blob/%r/%f#L%l")
+   (file-url-format           :initform "https://%h/%o/%n/-/blob/%r/%f#%l")
```

Just amend that change to the first commit.

lisp/forge-repo.el
tarsius: Thanks, I have just fixed that on master.

raszi:
> Edit: I just realized that this wasn't a typo per se, so ignore.
> 
> While working on a feature branch, it is best to not "preserve mistakes". E.g. it is completely irrelevant that you made a typo in the first commit and then fixed it like so in the second commit:

Well, if there are more people working on the same branch rebasing and force pushing could cause more harm than any good therefore I prefer to not change the past but to keep the commit history as it is. If it is really necessary then everything can be squashed into one commit when merging the PR.

raszi:
> > Merge branch 'master' into browse-file
> 
> Please rebase instead of merging. I have just done this now and then pushed the result to your branch.

Same as my other reply above. I don't believe rebasing and force pushing is a good practice when you are working with remote branches.

What I can do is to re-open the PR from a different branch without including the merge commit and limiting the number of commits I introduced but that can be also done by doing a squash commit.

What do you prefer?

tarsius:
Obviously we are on different sides of the rebase vs. merge discussion and there is no point in trying to convince each other here. This discussion has been had before and we can just go look up some careful presentations of the arguments of either position.

Just one thing; [IMO obviously] squashing is not a compromise, that's much much worse than either rebasing or merging. Using one branch per iteration is also awful. An acceptable compromise would be to delay history cleanup until right before the merge.

Ahh... I tried but I cannot simply let this go.

> I prefer to not change the past but to keep the commit history as it is

I prefer to not merge a commit that contains something so obviously fundamentally broken as this:

```lisp
(defun forge-browse-buffer-file ()
  "Visit the url corresponding to the actual file and location."
  (interactive
   (browse-url
    (let
        ((rev (magit-get-current-branch))
         (repo (forge-get-repository 'stub))
         (file (magit-file-relative-name buffer-file-name))
         (line-number (line-number-at-pos)))
      (forge--format repo 'file-url-format
                     `((?r . ,rev) (?f . ,file) (?l . ,line-number)))))))
```

There is absolutely no way I would ever knowingly merge something like this into one of my projects.

Ultimately this is my project and I decide how we do things around these parts. (Sorry for being harsh in my response. Your refusal to use the workflow that this project uses rubbed me the wrong way and I am just not in the mood to moderate myself today.)

Also, I actually started working on this myself after I looked at your implementation and did not like it. So there's really nothing left for us to do here and I am closing this.

raszi:
Yes, I agree obviously we are on different sides on this, and I have no bad feelings about that.

I offered to create a separate branch and fix the commits as you suggested or as an alternative merge the changes and squash the commit into once, I asked you to decide on this question.

I believe I was flexible on that front even though we have a disagreement how git and feature branches should be used.

I was willing to work on it even further since this is my first elisp pull-request ever and I guess it is far from being perfect but since you are implementing it and it will end up in the upstream I am satisfied with the outcome.

Lenbok:
> Also, I actually started working on this myself after I looked at your implementation and did not like it. So there's really nothing left for us to do here and I am closing this.

@tarsius Did this feature get implemented in the end?

PR #323, opened 2020-12-14
Create variables for custom predefined PR messages

KaratasFurkan:
We are using a pattern for PR messages in the company I work which is: "TARGET - #TICKET_NO MESSAGE"

For example: "STAGING - #9876 Create new variable"

Also adding a prefix like "FEATURE: " or "WIP: " or "BUGFIX: " may be common in some codebases IMO.

With this PR, the message template for `forge-create-pullreq` will become: "# **forge-create-pullreq-message-prefix** **forge-create-pullreq-message-number-prefix** |**BRANCH_NAME** ("|" means cursor position)

Example1:
With this setup:
```
(setq forge-create-pullreq-message-prefix "STAGING -")
(setq forge-create-pullreq-message-number-prefix t)
;; Lets say branch name is "9876-add-user-model" and last commit message is "Create User model" 
```
Auto generated PR message is:
```
STAGING - #9876 Create User model
```

Example2:
With this setup:
```
(setq forge-create-pullreq-message-prefix "FEATURE: ")
(setq forge-create-pullreq-message-number-prefix t)
;; Lets say branch name is "add-user-model" (notice that there is no number) and last commit message is "Create User model" 
```
Auto generated PR message is:
```
FEATURE: Create User model
```

I find this useful but it may be kinda opinionated, the decision is yours.

Thanks for this great package, btw.

### DIFF
--- a/lisp/forge-post.el
+++ b/lisp/forge-post.el
@@ -36,6 +36,15 @@
   :options '(visual-line-mode
              turn-on-flyspell))
 
+(defcustom forge-create-pullreq-message-prefix ""
+  "Prefix for new pull request message."
+  :type 'string)
+
+(defcustom forge-create-pullreq-message-number-prefix nil
+  "Insert issue/ticket/task number to new pull request message if
+source branch starts with a number."
+  :type 'string)
+
 ;;; Class
 
 (defclass forge-post (forge-object) () :abstract t)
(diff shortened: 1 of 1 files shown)


tarsius:
The existing code that this touches is rather wacky and I would rather not complicate it further before taking care of that. So I probably won't merge this or a later iteration before next year.

Regardless of that complication and the delay that comes with it, I think this is to specific to your particular use-case. This should probably be implemented as some `forge-...-function` variable and all the logic that is specific to how you do it should be inside your own function.

KaratasFurkan:
> I think this is to specific to your particular use-case
I agree.

> This should probably be implemented as some forge-...-function variable and all the logic that is specific to how you do it should be inside your own function.

This make sense, lets close this PR now, I will reopen it if I implemented it as you recommend.

Thank you.

PR #326, opened 2020-12-27
Add support for gitee.com

loyalpartner:
Your software is very good, I like it very much, but in my country we work with gitee.com, I made a little change to open the commit and issues in gitee.com in emacs, I hope you can pass this pr, so that more people can use this little feature easily.

### DIFF
--- a/lisp/forge-core.el
+++ b/lisp/forge-core.el
@@ -65,6 +65,8 @@
      "code.orgmode.org" forge-gogs-repository)
     ("bitbucket.org" "api.bitbucket.org/2.0"
      "bitbucket.org" forge-bitbucket-repository)
+    ("gitee.com" nil
+     "gitee.com" forge-gitee-repository)
     ;; Semi-Forges
     ("git.savannah.gnu.org" nil
      "git.savannah.gnu.org" forge-cgit**-repository)
--- a/lisp/forge-gitee.el
+++ b/lisp/forge-gitee.el
@@ -1,0 +1,38 @@
+;;; forge-gitee.el --- Gitee support              -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018-2020  Jonas Bernoulli
+
+;; Author: Jonas Bernoulli <jonas@bernoul.li>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; Forge is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; Forge is distributed in the hope that it will be useful, but WITHOUT
+;; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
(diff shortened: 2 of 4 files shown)


lisp/forge-core.el
```diff
@@ -75,7 +75,10 @@
     ("git.suckless.org" nil
      "git.suckless.org" forge-stagit-repository)
     ("git.sr.ht" nil
-     "git.sr.ht" forge-srht-repository))
+     "git.sr.ht" forge-srht-repository)
+    ;; gitee (API unsupported)
```
tarsius, line 79: Put this below the bitbucket entry into the "Forges (API unsupported)" group. (Don't add an additional comment, the comment I just mentioned covers that.)

lisp/forge-gitee.el
```diff
@@ -0,0 +1,38 @@
+;;; forge-bitbucket.el --- Bitbucket support      -*- lexical-binding: t -*-
```
tarsius, line 1: You forgot to adjust this. When you edit this make sure `-*- lexical-binding: t -*-` begins at the same column as before.

tarsius (changes requested):
Please update the "Partially Supported Forges" node in the manual. (Only edit `forge.org`; I will regenerate `forge.texi`.)

lisp/forge-core.el
```diff
@@ -75,7 +75,10 @@
...
+    ("gitee.com" nil
+     "gitee.com" forge-gitee-repository))
```
tarsius, line 81: Put this below the bitbucket entry into the "Forges (API unsupported)" group. (Don't add an additional comment, the comment I just mentioned covers that.)

loyalpartner:
Thank you for your guidance, I have changed according to what you said。

tarsius:
You omitted many of the slots that are set for many of the other classes.  Why?

loyalpartner:
i use M-x "forge-" to found the relation with them, but i can't found how to use them ,so i omit it.

tarsius:
> i use M-x "forge-" to found the relation with them, but i can't found how to use them ,so i omit it.

Some or even all of those are indeed unused. At present. I have added these slots because I expect to be using them eventually. Once that happens the values will already be ready to use for all forges that are currently supported. It would be nice if that also applied to gitee.com, especially because I don't use that myself, which means it would be more work for me to do it then than it would be for you to do it now.

tarsius:
I notice you added the suggested commit summary when squashing the commits. Unfortunately you also added the period at the end of the summary back again. Please remove that like for all other commits in this repository.

tarsius:
Please let me know when you have added the missing slots, then we can still add this.

tarsius:
I have added [a faq entry](https://github.com/magit/forge/wiki/FAQ#do-you-plan-to-addextend-support-forge-x) about the forges I plan or don't plan to support. Because I do not plan to implement support for this particular forge I am closing this issue. That doesn't mean that you cannot comment here or shouldn't discourage you from contributing support for favorite forge yourself.

PR #333, opened 2021-03-01
docs/forge Add :ensure t for use-package

Raphexion:
For someone new to emacs/magit/forge it probably helpful to include
the necessary :ensure t, even if that is not the main point of the
code snippet.


tarsius:
I disagree and think that because it has nothing to do with the point of the snippet, adding it is actually confusing.

PR #346, opened 2021-03-31
manual: Fix typo creating issues

miguelbernadi:
The "Creating topics" section wrongly claimed that the command called
with C-c C-n in the Issues section was forge-create-pullreq when it's
actually forge-create-issue. Fix the invoked command.


tarsius:
Thanks!

Issue #328, opened 2021-01-14
forge-pull silently fails for gitlab


ziprandom:
I'm using magit-forge for private repositories of an organization on gitlab.com and have to pin forge at 269effb4954071f590954c0e5b4faba3f27d03d0 (i haven't figured out whether this is really the last working commit). 

For current versions of forge I can't sync with the repository anymore. calling `forge-pull` will start the process, says `pulling issues` for an instant before doing the git pull and finish without an error and without showing any issues or pull requests. 

This also happens when I delete the database and start from scratch. I'm not sure whether this might be related to authentication, but the tokens & setup work flawless for the earlier version. 

There's no error, nothing in `*Messages*`, just no output in the magit ui. 

Using the same version of forge fit a github repository works so I assume this is related to the gitlab connector.

tarsius:
> says `pulling issues` for an instant

What does it say exactly? It should say something like `Pulling foo/bar issues`. If it says just `Pulling  issues`, then that it at least narrows down the issue a bit. Otherwise make sure that `foo/bar` actually identifies the upstream repository and not your fork for example.

ziprandom:
@tarsius thanks for your quick reply!

```
Pulling <org>/<repo> done
Storing <org>/<repo>...done
Running git fetch origin
Git finished
```

and shows

```
Stashes (8)...
Recent commits...
<-- end of buffer -->
```

tarsius:
Maybe there aren't actually any open issues and pull-requests?

ziprandom:
I'm comparing againste the same repository using the sam api token 

The only difference is the version of forge. The old version shows plenty of Issues/PR's.

tarsius:
I suggest you sprinkle debug statements to find out where things start going south.

tarsius:
Does `M-x forge-list-topics` list all the topics that it should list?

ziprandom:
`M-x forge-list-topics` doesn't list any topics. 

I also tried with `HOME=$(mktemp -d) emacs -q` to make sure no effects from the rest of my config occur. 

From what I experience it should not work for any gitlab user. The fact that I'm the only one to report it indicates it's only my though problem ..

tarsius:
... yeah, I can't really think any tip beyond debug statements to figure out what (if any) steps are unexpectedly omitted.

tarsius:
Please check if 18f03a743739e30dbcdfcae663d03c3dd24278e6 fixes this.

ziprandom:
it does indeed!

Issue #289, opened 2020-07-24
forge-pull self-hosted GitLab instance issues


tru2dagame:
Hi,
I am able to run `forge-pull` with my self-hosted GitLab instance with no errors. But it only add the `repository` to sqlite database.

![image](https://user-images.githubusercontent.com/1673006/88417093-ddda6780-ce13-11ea-9b77-3d04cd364c9a.png)

![image](https://user-images.githubusercontent.com/1673006/88417008-c3a08980-ce13-11ea-918c-45c928ebee5c.png)

The token and setup works with the GitLab api because I am able to create issues with forge. But can't see it in forge.

tarsius:
And with gitlab.com and github.com it works as expected?

tru2dagame:
Hi @tarsius, I haven't tried gitlab.com.
And yes, github.com and I also have a GitHub Enterprise instance, they do works as expected.

![image](https://user-images.githubusercontent.com/1673006/88450315-a6ee6b00-ce80-11ea-8cad-13d25f1309e6.png)


![image](https://user-images.githubusercontent.com/1673006/88450274-5c6cee80-ce80-11ea-89d1-d77a9f6877f4.png)

tru2dagame:
@tarsius just tried gitlab.com. It doesn't work for me.

tarsius:
I'm afraid I have no idea what might be wrong.

tru2dagame:
Hi @tarsius As the GitLab api has already support graphql api, any chance we could upgrade the GitLab to use that?

tarsius:
When did *that* happen?

tru2dagame:
@tarsius 

> https://docs.gitlab.com/ee/api/graphql/

According to the docs, I think that start it from version 11.0 and always enabled in GitLab 12.1.

Now the latest version of GitLab is 13.3.

tarsius:
Some more meta information:

- https://docs.gitlab.com/ee/development/fe_guide/graphql.html
- https://docs.gitlab.com/ee/api/README.html#road-to-graphql

> According to the docs, I think that start it from version 11.0 and always enabled in GitLab 12.1.

Yeah, stumbled into a sentence that implies that too. But so far no luck locating a release announcement that talks about it. Neither of these two release announcements do as far as I can tell. Ah well.

tarsius:
So anyway... I probably won't work on it soon.

tru2dagame:
Thanks. Got it.

[...]

PR #347, opened 2021-04-02
Correctly pull GitLab data with forge.omitExpensive

lvanengelen:
When `forge.omitExpensive` is set, some expensive fetches are avoided by
adding empty entries for them in a result variable.  However, this broke
handling the result of fetching the repository data: it was added a
level deeper than expected in the remaining code.

Fix this issue by looking at the result data: if the first element is a
cons cell it is the repository data and should be appended, otherwise it
is an additional fetch (issues, pull requests, ...) and should be pushed
to the results variable.

### DIFF
--- a/lisp/forge-gitlab.el
+++ b/lisp/forge-gitlab.el
@@ -51,7 +51,10 @@
                   (dir default-directory)
                   (val nil))
               (lambda (cb &optional v)
-                (when v (if val (push v val) (setq val v)))
+                (when v
+                  (if (consp (car v))
+                      (setq val (append v val))
+                    (push v val)))
                 (let-alist val
                   (cond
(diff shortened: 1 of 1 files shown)


lvanengelen:
I ran into this issue when trying to pull data from a large repository. I am only a Lisp beginner: the check might not be the nicest but it was the simplest I could think of. After a quick look at the open issues I thought that #289 and #328 might be related.

tarsius:
Nice find.  Thanks!

> Fix this issue by looking at the result data ...

I decided to instead just not mess up the data in the first place. :P

> After a quick look at the open issues I thought that #289 and #328 might be related.

Might be, might be. I certainly hope so.

PR #373, opened 2021-06-24
Add yaml dependency and use it for parsing YAML

zkry:
This PR adds the yaml dependency located at https://github.com/zkry/yaml.el/ (https://melpa.org/#/yaml). Using a specialized YAML parser for this simplifies the code and makes it more resilient to syntactical variations that may occur. This dependency may also ease the extraction of different metadata later on.

### DIFF
--- a/default.mk
+++ b/default.mk
@@ -29,6 +29,7 @@ DEPS += markdown-mode
 DEPS += transient/lisp
 DEPS += treepy
 DEPS += with-editor
+DEPS += yaml
 
 VERSION ?= $(shell test -e $(TOP).git && git describe --tags --abbrev=0 | cut -c2-)
 
--- a/forge-pkg.el
+++ b/forge-pkg.el
@@ -8,6 +8,7 @@
     (let-alist      "1.0.6")
     (magit          "3.0.0")
     (markdown-mode  "2.4")
-    (transient      "0.3.3"))
+    (transient      "0.3.3")
+    (yaml           "0.3.3"))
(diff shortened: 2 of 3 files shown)


tarsius:
Thanks!

Before I merge this, could you please address the following issue.

```
lib/yaml/yaml.el:636:1: Warning: defconst `yaml--terminal-rules' docstring
    wider than 80 characters
```

By the way, I am unable to parse that sentence anyway:

> List of rules that indicate that the parse tree should end at and children ignored.

The next doc-string (of `yaml--walk-events`) is also too long. Not sure why the byte-compiler doesn't complain about that one, but you should shorten it anyway.

Speaking of long lines...
I would like to encourage you to stay below 80 characters. If the occasional line is up to 95 characters long, then that is also still okay. But that line (2106) that is 197 characters lines long without there being any benefit whatsoever to that, that's ... not so nice. And 150, 120 and even "just" 100 are also rather problematic.

I strongly urge you to stay around 80 with the occasional 95 overflow exception.

To ease that you might want to add `(declare (indent 2))` to `yaml--rep2`, and declarations for other functions similar.

zkry:
Thanks for the suggestions! I updated the yaml package so that lines don't exceed 80 characters. Yeah, that docstring was poorly written. I updated that as well (those were the grammar rules that were recursive on text, going really deep causing an error sometimes).

tarsius:
Great. Instantly looks better. :grin:

I intend to merge this with the above fixup changes.

PR #376, opened 2021-07-06
Add values to topic-title and topic-state sections

nagy:
This makes it a bit easier to extract these values.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -461,7 +461,7 @@ identifier."
 
 (cl-defun forge-insert-topic-title
     (&optional (topic forge-buffer-topic))
-  (magit-insert-section (topic-title)
+  (magit-insert-section (topic-title (oref topic title))
     (insert (format "%-11s" "Title: ") (oref topic title) "\n")))
 
 (defvar forge-topic-state-section-map
@@ -471,7 +471,7 @@ identifier."
 
 (cl-defun forge-insert-topic-state
     (&optional (topic forge-buffer-topic))
-  (magit-insert-section (topic-state)
+  (magit-insert-section (topic-state (oref topic state))
     (insert (format "%-11s" "State: ")
             (symbol-name (oref topic state))
             "\n")))


tarsius:
What do you need that for? Could you do it the hard way instead, for the time being?

I cannot help but notice that you only do it for some of the headers. I assume that's the ones you need. As the maintainer I would have to decide if I wanted to do this for none or all and if it is the latter whether it needs to be somehow prettified, because boilerplate.

Okay now, football!

tarsius:
Looks like you want to use `(oref (magit-current-section) title)`.
But you should just use `(oref forge-buffer-topic title)`, which has the advantage that it works anywhere in the buffer.

nagy:
First of all, your extensions are very helpful to me so I am very thankful for that.

> What do you need that for?

More and more packages start to use `magit-section` to render their contents.
This has prompted me to come up with some key bindings and functions to deal
with the provided functionality of `magit-section` in my personal config. For
example I can copy a value of a section to the kill-ring like this:

```elisp
(defun kill-magit-section-value ()
  "Kill the value of the magit-section at point."
  (interactive)
  (let ((value (oref (magit-current-section) value)))
    (message (kill-new (format "%s" value)))))
(map! "H-q y" #'kill-magit-section-value)
```

`map!` is a doom-emacs helper btw. Without this I would have to manually skip
past the "Title:" prefix of a topic title for example to achive the same effect.
But the greater value I see in this is that, when I know that a certain project
is using `magit-section`, I can reuse this same functionality and key-bindings
across all of them.

Regarding any string based value, lets say for example the title of a topic, I
can search all open buffers for this text:

```elisp
(defun magit-section-buffer-search-value ()
  (interactive)
  (swiper-all (format "%s" (oref (magit-current-section) value))))
(map! "H-q B" #'magit-section-buffer-search-value)
```

Now with a press of `H-q B` it opens a list of all open buffers that contain
this topic title. This can help me to find the buffer of an email or an org-mode
buffer where I have some notes about this topic.

Another one is value of the state. This is not a string, but a symbol.
Magit-section also give access to the lineage. For the state it is
`(topic-state topic-title topicbuf)`. I could write a function that should show
me all open buffers where there is a section where the last child contains
exactly this value at point, lets say `'open'`. In other words show me all open
buffers which show an open (meaning unclosed) issue.

> I cannot help but notice that you only do it for some of the headers. I assume
> that's the ones you need.

Well, these two were just the only straight forward ones that I could acutely
find. For example regarding the "Label:" , if a certain topic has two labels
"bug" and "help wanted", would I rather have the value of the section be a comma
separated string or rather an actual lisp list like `(list "bug" "help wanted")`?
I would probably go for the latter but I still want to think through possible
use cases of each variant.

If the value of a section gets more complicated, I could see the following function useful as well:

```elisp
(defun kill-magit-section-value-as-json ()
  (interactive)
  (let ((value (json-encode (oref (magit-current-section) value))))
    (message (kill-new (format "%s" value)))))
```

This serializes the section value into a json value and therefore the above
example would kill this into the kill-ring: `["bug","help wanted"]`. I could see
this being useful if you want to paste this into a javascript file or prompt for
example. A further down the line use case could be that, you could have two
issues open and want to transfer ( i.e. copy ) the label from one issue to the
other. You could yank the raw label value like this and paste them onto the
other issue. But for this the setter would need to understand how to apply these
values and modify the Labels.

> whether it needs to be somehow prettified, because boilerplate.

Do you mean the repeated use of `#'oref'` ? Extract that out into a function ?
I can gladly work on the rest of the headers if you see any of this useful.

> Okay now, football!
👀

nagy:
> Looks like you want to use `(oref (magit-current-section) title)`.
> But you should just use `(oref forge-buffer-topic title)`, which has the advantage that it works anywhere in the buffer.

Thank you for the hint

nagy:
In case somebody stumbles upon this and still wants to have that desired effect, you can put this into your config:

```elisp
(defun +forge-set-topic-title-value ()
  (save-excursion
    (backward-char)
    (oset (magit-current-section) value (oref forge-buffer-topic title))))
(advice-add  #'forge-insert-topic-title :after #'+forge-set-topic-title-value)

(defun +forge-set-topic-state-value ()
  (save-excursion
    (backward-char)
    (oset (magit-current-section) value (oref forge-buffer-topic state))))
(advice-add  #'forge-insert-topic-state :after #'+forge-set-topic-state-value)
```

tarsius:
I understand where you are coming from but we always have to strike a balance between making all known and unknown potential third-party uses as easy to accomplish as possible and not drowning in a see of possibilities. I think what we currently do well in this regard. The implied convention is this: if the package that defines the section does need to use the sections value then set the value, otherwise other things probably won't need the value either, so don't.

I would recommend against using such advice for the benefit of `kill-magit-section-value`. If you want such a command then I would rather recommend to change it to take into account that there will always be sections that don't define the value and fallback to kill the first line of the section in such cases. Sure, at times that won't give you exactly what you want, then again depending on where you later yank the killed string the section value won't always the right value either.

PR #381, opened 2021-07-24
Warn on restrictive url-privacy-level settings that prevent using forge-github

akater:
It might be better to warn already when loading ghub but file description does not specify it as GitHub-specific so I hesitated.

It took me at least a year to start using forge, and this was one of the issues which is why I think emtting a clear warning is beneficial for users.

Debugger only shows comparatively cryptic “Request forbidden by administrative rules”.  Details about User-Agent are hidden by default.  Some users might not even know how to look further; others might be scared of the debugger.  Overall, it's easy to abandon or delay using forge for no good reason in such situation.

### DIFF
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -745,6 +745,42 @@
                :noerror noerror :reader reader
                :callback callback :errorback errorback))
 
+(unless (fboundp 'forge-github--check-url-privacy)
+  (let (warned-once)
+    (defun forge-github--check-url-privacy ()
+      "Check whether `url-privacy-level' settings prevent using forge-github.
+Return non-nil value describing the issue if that's the case.
+
+Display warning unless already displayed."
+      (when (featurep 'url-privacy)
+        (cl-macrolet ((or--return-form
+                       (&rest forms)
+                       `(cond
+                         ,@(mapcar (lambda (form)
+                                     `(,form
+                                       ',(if (and (eq 'and (car form))
+                                                  (cdr form))
+                                             (car (last form))
+                                           form)))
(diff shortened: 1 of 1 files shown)


tarsius:
> Debugger only shows comparatively cryptic “Request forbidden by administrative rules”.

The error is: `error in process sentinel: HTTP Error: 403, "Forbidden", "/graphql", ((message . "Request forbidden by administrative rules. Please make sure your request has a User-Agent header (http://developer.github.com/v3/#user-agent-required). Check https://developer.github.com for other possible causes.") (documentation_url . "https://github.com/magit/ghub/wiki/Github-Errors"))`.

Which I consider a good error message. Before I saw the actual error I came up with this error message: `Github doesn't work when the User-Agent is not send.  Please adjust 'url-privacy-level'.`, which is a little bit better for Emacs users because it mentions `url-privacy-level`. However upon seeing the first error a user who has adjusted `url-privacy-level` will hopefully remember that the customized that at some earlier time (not necessarily the name, but that they have to adjust some `url` setting). If not, then a search engine should do the rest, especially now that it takes them here.

IMO it is overkill to add the second warning, which I implemented by calling this function from `ghub--retrieve` and `ghub--graphql-retrieve`:

```elisp
(defun ghub--assert-user-agent (forge)
  (when (and (memq forge '(github nil))
             (if (listp url-privacy-level)
                 (memq 'agent url-privacy-level)
               (eq url-privacy-level 'paranoid)))
    (error "Github doesn't work when the User-Agent is not send.  %s `%s'."
           "Please adjust" 'url-privacy-level)))
```

PR #383, opened 2021-08-06
Silence compiler warning from Emacs without X support

akater:


tarsius:
Such declarations are no-opts if the variable is already defined, so I dropped the `(unless (boundp 'x-stretch-cursor) ...)`.

PR #419, opened 2021-10-04
docs: Add a FAQ: HTTP Error 502 "Bad gateway"

yashi:
Many of us get this 502 Bad gateway error while `forge-pull` (f y).  Add a link to the related discussion.

ps. `ox-texinfo` didn't allow me to use the exact notation:

    HTTP Error: 502, "Bad gateway", "/graphql"

Thus, the FAQ is without punctuation.


tarsius:
Thanks. I've added my own FAQ entry about this, which I believe to be more accurate.

PR #423, opened 2021-10-05
Add Gnome's gitlab instance to forge-alist

tsdh:
This PR adds the Gnome project's GitLab instance to `forge-alist`.

### DIFF
--- a/lisp/forge-core.el
+++ b/lisp/forge-core.el
@@ -61,6 +61,8 @@
      "salsa.debian.org" forge-gitlab-repository)
     ("framagit.org" "framagit.org/api/v4"
      "framagit.org" forge-gitlab-repository)
+    ("gitlab.gnome.org" "gitlab.gnome.org/api/v4"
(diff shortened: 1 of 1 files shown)


lisp/forge-core.el
```diff
@@ -61,6 +61,8 @@
      "salsa.debian.org" forge-gitlab-repository)
     ("framagit.org" "framagit.org/api/v4"
      "framagit.org" forge-gitlab-repository)
+    ("gitlab.gnome.org" "gitlab.gnome.org/api/v4"
+     "GNOME-GitLab" forge-gitlab-repository)
```
tarsius, line 65: So far we have always used the same value for `id` as for `githost`. I think I want to stick to that.

tsdh: Allright, I've changed that accordingly in the recently pushed version of the commit.

tarsius:
Thanks!

PR #431, opened 2021-10-26
WIP - Super initial gitea notification support

MorganJamesSmith:
Hey,

I was looking to use forge for gitea but there doesn't seem to be any support for, well, anything at all.

I tried my best for the past 2.5 hours and came up with this. It successfully lists notifications. It's mostly copy paste from the github version and it was quite challenging because I'd never seen these generics or eieio stuff in elisp before.

So I mostly wanted to start a discourse here to discuss the current state of gitea support.

Is anyone currently working on this? This would be a pretty big thing for me to try and tackle so I'd love it if someone was.

If no-one is working on it, I'd definitely need some serious help :P. I'm a master of rebasing local changes so please feel free to simply throw some commits on this branch to show me how it's done. I'd really appreciate that.

The manual says gitea is the next forge that will supported but the manual also has a bunch of wrong keybindings so I'm not sure if it's up to date (that was fun thinking my install was wrong because there was no "f y" option. Or is my install actually broken? Should I see an "f y" option?).

P.S. I've never actually used forge (other than viewing a half broken gitea notification page). I should probably try it out before committing to this...

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -164,7 +164,7 @@ If pulling is too slow, then also consider setting the Git variable
   (interactive)
   (let* ((repo  (forge-get-repository 'stub))
          (class (eieio-object-class repo)))
-    (if (eq class 'forge-github-repository)
+    (if (memq class '(forge-github-repository forge-gitea-repository))
         (forge--pull-notifications class (oref repo githost))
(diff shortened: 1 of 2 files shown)


tarsius:
I made some progress on this. I'll try to resume working on it later this week.

PR #445, opened 2021-12-10
Fix forge-pull-topic can not pull topic

chenyanming:
It is necessary to call `forge-get-topic` to get the topic.


tarsius:
Since 8264234db61b8a7f427b972aaef6235f9f6a13fa we avoid using objects as interactive arguments.

PR #447, opened 2021-12-15
Get and show avatar in `forge-topic-mode`

chenyanming:
This pull request can get and download avatar from github and shows them in `forge-topic-mode`.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -96,6 +96,12 @@ This variable has to be customized before `forge' is loaded."
              magit-mode-hook)
   :type '(list :convert-widget custom-hook-convert-widget))
 
+(defcustom forge-avatar-dir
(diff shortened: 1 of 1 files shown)


chenyanming:
image

tarsius:
Related https://github.com/magit/magit/issues/4414.

tarsius:
That serves well as a proof-of-concept, but a lot more would have to be done until I would be comfortable adding that to forge.

I have added some stubs to forge so you can implement this for your own use without having to touch forge code. Maybe you could turn that into a package over time or we could merge it into forge once you have taken care of all the details.

A first issue is that forge not only supports github. Then we would like to fetch asynchronously but display the image once it has arrived without requiring an explicit refresh by the user...

See the gravatar support in magit and also the work done following https://github.com/magit/magit/issues/4414.

chenyanming:
Thanks！The `forge-format-avatar-function` is sufficient for me！Much better than `add-advice`.

PR #448, opened 2021-12-23
Only enable bug-reference-mode when Forge repository data is available 

leotaku:
This PR changes `forge-bug-reference-setup` to only enable `bug-reference-mode` or `bug-reference-prog-mode` when Forge repository data is available.

According to the [function doc-string](https://github.com/magit/forge/blob/9c3b53cdf6c0aae18455a8b9f573846c43ac9df1/lisp/forge-topic.el#L912), this is already the expected behavior, and the current behavior of always enabling `bug-reference-mode` or `bug-reference-prog-mode` is a bug.


tarsius:
Thanks!

PR #451, opened 2021-12-28
Reformat transient menu items for consistency

nbarrientos:
Just a cosmetic change to make the `magit-status-jump` transient look more consistent and elegant (capitalised and using space as word separator).

Before:

image

After:

image

The patch makes the displayed options consistent as well with the titles of the sections in the Magit status buffer:

image


tarsius:
Thanks!

PR #485, opened 2022-04-15
Correct key name

nbarrientos:
This function is meant to set to nil the target project in case of an error bringing the information about the project itself, not the source project.


tarsius:
Thanks!

PR #488, opened 2022-04-25
Visit the issue/pullreq at point in short link notation

### LINKED ISSUE #43: "No such file or directory" when opening buffer with nonexisting path
Attempting to create a buffer with a path with non-existing directory under a git repository will result in a "No such file or directory" error while `magit-forge` is installed.

Stack trace:

```
Debugger entered--Lisp error: (file-missing "Setting current directory" "No such file or directory" "/home/vladimir/tmp/2018-12-27-scratch/17:01:41/")
  call-process("git" nil (t nil) nil "--no-pager" "--literal-pathspecs" "-c" "core.preloadindex=true" "-c" "log.showSignature=false" "remote")
  apply(call-process "git" nil (t nil) nil ("--no-pager" "--literal-pathspecs" "-c" "core.preloadindex=true" "-c" "log.showSignature=false" "remote"))
  process-file("git" nil (t nil) nil "--no-pager" "--literal-pathspecs" "-c" "core.preloadindex=true" "-c" "log.showSignature=false" "remote")
  apply(process-file "git" nil (t nil) nil ("--no-pager" "--literal-pathspecs" "-c" "core.preloadindex=true" "-c" "log.showSignature=false" "remote"))
  (let ((process-environment (magit-process-environment)) (default-process-coding-system (magit--process-coding-system))) (apply (function process-file) process infile buffer display args))
  magit-process-file("git" nil (t nil) nil "--no-pager" "--literal-pathspecs" "-c" "core.preloadindex=true" "-c" "log.showSignature=false" "remote")
  apply(magit-process-file "git" nil (t nil) nil ("--no-pager" "--literal-pathspecs" "-c" "core.preloadindex=true" "-c" "log.showSignature=false" "remote"))
  (if magit-git-debug (let (log) (unwind-protect (progn (setq log (make-temp-file "magit-stderr")) (delete-file log) (let ((exit (apply (function magit-process-file) magit-git-executable nil (list t log) nil args))) (if (> exit 0) (progn (let ((msg "Git failed")) (if (file-exists-p log) (progn (setq msg (let ((temp-buffer (generate-new-buffer " *temp*"))) (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (insert-file-contents log) (goto-char (point-max)) (cond ((functionp magit-git-debug) (funcall magit-git-debug (buffer-string))) ((run-hook-wrapped 'magit-process-error-message-regexps (function (lambda (re) (re-search-backward re nil t)))) (match-string-no-properties 1)))) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))))) (let ((magit-git-debug nil)) (save-current-buffer (set-buffer (magit-process-buffer t)) (magit-process-insert-section default-directory magit-git-executable args exit log))))) (message "%s" msg)))) exit)) (condition-case nil (progn (delete-file log)) (error nil)))) (apply (function magit-process-file) magit-git-executable nil (list t nil) nil args))
  magit-git-insert("remote")
  apply(magit-git-insert "remote")
  (progn (apply (function magit-git-insert) args) (split-string (buffer-string) "\n" t))
  (unwind-protect (progn (apply (function magit-git-insert) args) (split-string (buffer-string) "\n" t)) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))
  (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (apply (function magit-git-insert) args) (split-string (buffer-string) "\n" t)) (and (buffer-name temp-buffer) (kill-buffer temp-buffer))))
  (let ((temp-buffer (generate-new-buffer " *temp*"))) (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (apply (function magit-git-insert) args) (split-string (buffer-string) "\n" t)) (and (buffer-name temp-buffer) (kill-buffer temp-buffer)))))
  magit-git-lines("remote")
  magit-list-remotes()
  #f(compiled-function (demand &optional remote) "Return the forge repository for the current Git repository." #<bytecode 0x21e9559>)(full)
  apply(#f(compiled-function (demand &optional remote) "Return the forge repository for the current Git repository." #<bytecode 0x21e9559>) full nil)
  forge-get-repository(full)
  forge-bug-reference-setup()
  run-hooks(find-file-hook)
  after-find-file(t t)
  find-file-noselect-1(#<buffer scratch.org> "~/tmp/2018-12-27-scratch/17:01:41/scratch.org" nil nil "~/tmp/2018-12-27-scratch/17:01:41/scratch.org" nil)
  find-file-noselect("~/tmp/2018-12-27-scratch/17:01:41/scratch.org" nil nil nil)
  find-file("~/tmp/2018-12-27-scratch/17:01:41/scratch.org")
  (progn (progn (fset 'sit-for (function (lambda (seconds &optional nodisp) (let ((this-fn saved-func-sit-for))))))) (find-file (concat (format-time-string "~/tmp/%Y-%m-%d-scratch/%H:%M:%S/" (current-time)) filename)) (if contents (progn (insert contents) (goto-char 1) (search-forward "XXX") (forward-char -3) (delete-char 3))))
  (unwind-protect (progn (progn (fset 'sit-for (function (lambda (seconds &optional nodisp) (let ((this-fn saved-func-sit-for))))))) (find-file (concat (format-time-string "~/tmp/%Y-%m-%d-scratch/%H:%M:%S/" (current-time)) filename)) (if contents (progn (insert contents) (goto-char 1) (search-forward "XXX") (forward-char -3) (delete-char 3)))) (progn (if (eq (symbol-function 'noflet|base) saved-func-sit-for) (fmakunbound 'sit-for) (fset 'sit-for saved-func-sit-for))))
  (let ((saved-func-sit-for (condition-case err (symbol-function 'sit-for) (void-function (symbol-function 'noflet|base))))) (unwind-protect (progn (progn (fset 'sit-for (function (lambda (seconds &optional nodisp) (let ((this-fn saved-func-sit-for))))))) (find-file (concat (format-time-string "~/tmp/%Y-%m-%d-scratch/%H:%M:%S/" (current-time)) filename)) (if contents (progn (insert contents) (goto-char 1) (search-forward "XXX") (forward-char -3) (delete-char 3)))) (progn (if (eq (symbol-function 'noflet|base) saved-func-sit-for) (fmakunbound 'sit-for) (fset 'sit-for saved-func-sit-for)))))
  cs/create-new-buffer("scratch.org")
  (lambda nil (interactive) (cs/create-new-buffer "scratch.org"))()
  funcall-interactively((lambda nil (interactive) (cs/create-new-buffer "scratch.org")))
  call-interactively((lambda nil (interactive) (cs/create-new-buffer "scratch.org")) nil nil)
  command-execute((lambda nil (interactive) (cs/create-new-buffer "scratch.org")))
```

It looks like the error is due to `magit-list-remotes` etc. are being called with the current buffer being the file buffer, whereas normally (i.e. in magit) they are called with `magit-status` etc. buffers.

nbarrientos:
I find it useful to quickly jump to an issue when for instance reviewing a pull request. Let's say that the pull request description reads as follows:

``` 1c-enterprise
Fix a bug.

Closes #43
```

Placing the point in between `4` and `3` and running `M-x forge-visit-issue` directly visits the issue after this patch, instead of prompting for which issue to visit. The matching could be more sophisticated and only react to integers at point preceded by `#`.

Maybe this could be integrated into `forge-issue-at-point` instead so the discovery would be done by `forge-current-issue` so no modifications to `forge-read-issue` would be needed.

The same idea could be applied to `forge-read-pullreq`, I imagine. Happy to contribute a patch as part of the same pull request.

Haven't check if the manual needs updating. I'll gladly do so if the change request goes forward.

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -343,15 +343,19 @@ read a topic to visit instead."
 (defun forge-visit-pullreq (pullreq)
   "View the current pull-request in a separate buffer.
 If there is no current pull-request or with a prefix argument
-read a PULLREQ to visit instead."
+read a PULLREQ to visit instead. If point is looking at a pullreq
+reference with Gitlab/Github notation try to visit the pullreq
+with that number."
   (interactive (list (forge-read-pullreq "View pull-request" t)))
   (forge-visit (forge-get-pullreq pullreq)))
 
 ;;;###autoload
 (defun forge-visit-issue (issue)
   "Visit the current issue in a separate buffer.
-If there is no current issue or with a prefix argument
-read an ISSUE to visit instead."
+If there is no current issue or with a prefix argument read an
+ISSUE to visit instead. If point is looking at an issue reference
+with Gitlab/Github notation try to visit the issue with that
+number."
   (interactive (list (forge-read-issue "View issue" t)))
   (forge-visit (forge-get-issue issue)))
 
--- a/lisp/forge-issue.el
+++ b/lisp/forge-issue.el
@@ -128,6 +128,9 @@
 (cl-defmethod forge-get-url ((issue forge-issue))
   (forge--format issue 'issue-url-format))
 
+(defun forge--issue-by-forge-short-link-at-point ()
+  (forge--topic-by-forge-short-link-at-point '("#") #'forge-get-issue))
+
 ;;; Sections
 
 (defun forge-current-issue ()
@@ -146,7 +149,8 @@
         (cond ((forge-issue-p post)
                post)
               ((forge-issue-post-p post)
-               (forge-get-issue post))))))
+               (forge-get-issue post))))
+      (forge--issue-by-forge-short-link-at-point)))
 
 (defvar forge-issues-section-map
   (let ((map (make-sparse-keymap)))
--- a/lisp/forge-post.el
+++ b/lisp/forge-post.el
@@ -80,7 +80,9 @@
                  (n (magit-get "branch" branch "pullRequest")))
         (forge-get-pullreq (string-to-number n)))
       (and-let* ((rev (magit-commit-at-point)))
-        (forge--pullreq-from-rev rev))))
+        (forge--pullreq-from-rev rev))
+      (forge--issue-by-forge-short-link-at-point)
+      (forge--pullreq-by-forge-short-link-at-point)))
 
 (defun forge-current-topic ()
   (or (forge-topic-at-point)
--- a/lisp/forge-pullreq.el
+++ b/lisp/forge-pullreq.el
@@ -221,6 +221,9 @@ yourself, in which case you probably should not reset either.
 (cl-defmethod forge-get-url ((pullreq forge-pullreq))
   (forge--format pullreq 'pullreq-url-format))
 
+(defun forge--pullreq-by-forge-short-link-at-point ()
+  (forge--topic-by-forge-short-link-at-point '("#" "!") #'forge-get-pullreq))
+
 ;;; Sections
 
 (defun forge-current-pullreq ()
@@ -239,7 +242,8 @@ yourself, in which case you probably should not reset either.
         (cond ((forge-pullreq-p post)
                post)
               ((forge-pullreq-post-p post)
-               (forge-get-pullreq post))))))
+               (forge-get-pullreq post))))
+      (forge--pullreq-by-forge-short-link-at-point)))
 
 (defvar forge-pullreqs-section-map
   (let ((map (make-sparse-keymap)))
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -358,6 +358,18 @@ identifier."
      (format-time-string "%s" (parse-iso8601-time-string (oref topic created)))
      t)))
 
+(defun forge--topic-by-forge-short-link-at-point (known-prefixes finder)
+  "Finds a topic by forge-dependant short link around point.
+The topic number is expected to be a number prefixed by any of
+the elements in KNOWN-PREFIXES. If a reference is found, FINDER
+is called and a topic object is returned if available."
+  (and-let* ((number (number-at-point))
+             (prefix (buffer-substring-no-properties
+                      (- (match-beginning 0) 1)
+                      (match-beginning 0))))
+    (and (member prefix known-prefixes)
+         (funcall finder number))))
+
 ;;; Mode
 
 (defvar forge-topic-mode-map


tarsius:
You should be able to press `RET` in that location to invoke `bug-reference-push-button`, which doesn't require confirmation (though depending on what Emacs version you use it might require some setup).

If you loose interest in implementing anything else after learning about that (or figuring out how to get it to work), I understand. After all I don't plan to implement anything myself for that very reason.

> The matching could be more sophisticated and only react to integers at point preceded by `#`.

That's what we should do. Look at the `bug-reference`-related code. Also for Gitlab `!N` also has to be supported.

> Maybe this could be integrated into `forge-issue-at-point` instead so the discovery would be done by `forge-current-issue` so no modifications to `forge-read-issue` would be needed.

Yes, that's where it belongs I think.

The decision whether to use the issue at point with or without confirmation or with completion should be handled inside `magit-read-issue`. Actually that's already implemented (except for picking up the issue at point as default), see `magit-dwim-selection`.

nbarrientos:
Hi @tarsius,

Thanks for replying.

> You should be able to press `RET` in that location to invoke
> `bug-reference-push-button`, which doesn't require confirmation
> (though depending on what Emacs version you use it might require some
> setup).

`bug-reference-mode` is not unknown to me, but, correct me if I'm wrong, looking into the code it seems that the goal is always to ultimately call `browse-url` on the URL that's derived from the bug number. I don't see a way to call custom code instead so I can instead visit the issue in Forge. The idea behind this patch proposal has always been that, to browse the issue using Forge. Then, once I'm visiting the issue, I can always `N b t` if I want to see it in a browser.

> That's what we should do. Look at the `bug-reference`-related
> code. Also for Gitlab `!N` also has to be supported.

ACK, this is doable.

> 
> > Maybe this could be integrated into `forge-issue-at-point` instead
> > so the discovery would be done by `forge-current-issue` so no
> > modifications to `forge-read-issue` would be needed.
> 
> Yes, that's where it belongs I think.

I'm happy to refactor a bit and make the matching better. Let me know though if you're interested in reviewing such changes, I understand that this topic might not be the most exciting one :)

tarsius:
> `bug-reference-mode` is not unknown to me, but, correct me if I'm wrong, looking into the code it seems that the goal is always to ultimately call `browse-url` on the URL that's derived from the bug number. I don't see a way to call custom code instead so I can instead visit the issue in Forge.

Right for those links it does that. But for debbug issues it uses the `debbugs` package. Not sure what it does if that isn't installed. So it is customizable somehow. I haven't really looked into it lately but believe I wanted to take a closer look eventually so that `bug-reference-mode`, `debbugs` and `forge` better complement each other, while avoiding duplication. ... But it always worked good enough for me and there always were more pressing things.

> I'm happy to refactor a bit and make the matching better. Let me know though if you're interested in reviewing such changes, I understand that this topic might not be the most exciting one :)

Your last contribution was excellent and very welcome, so I would be happy to review whatever you implement.

I think pretty soon we will get to the point where we need *design decisions*™ and just making incremental improvements could get us into some dead-end. I would suggest that at least initially you concentrate on what you originally proposed and mostly keep things under control of `forge`. But do keep out an eye for conflicts with `bug-reference`, at least as far as key bindings and text properties go.

tarsius:
Ah I think I remember my past reasonsing:

1. Something else (`debbugs`) already does something useful when pressing `RET`.
2. It also hightlights the issue number.
3. So let it do those things and don't try to compete.
4. Because "visit this issue" already exits in *some* form, I can probably get away without implementing "visit issue in forge" for a while.
5. Hopefully those who want to visit in Forge will figure out the key binding to do that.
6. Hopefully they are also okay with not being able to click on the link to do so.

And then you showed up. :grinning:

nbarrientos:
Hi,

I totally understand your point about trying to achieve an elegant and reasonable integration with other built-in tools. It makes total sense of course to avoid reinventing the wheel.

I've applied some changes aiming at providing the functionality I'd like to have but leaving the way clear for `bug-reference` et al to do the job they're meant to do. In other words, the patch just shows `forge` how to interact with the buffer, without creating any conflicting overlay or keybinding. References to issues and pullreqs are not highlighted and the integration is friction-less and merely on the Forge side.

This should leave the door open to further integration with other tools (to be implemented in `forge` itself or by third-party configuration declared by the user).

The matching could be even better as things like `bug#34` will be picked up but it's *Good Enough* for me. Happy to improve it if you have any suggestion :smile:

tarsius:
Nice and tidy. Thanks!

I moved the location of the main function and switched from `when` to `and` because it's all about the return value, not side-effects.

PR #494, opened 2022-05-08
Fix `:initform` form for sqlite-builtin and sqlite-module backends

marsam:
ab637b7 accidentally set `epkg-package` as :initform to the sqlite-builtin and sqlite-module backends.


tarsius:
Thanks!

PR #504, opened 2022-07-19
Add link to autofill GitHub token creation form

jidicula:
The GitHub token creation form can be autofilled with scope
selections:
https://github.com/settings/tokens/new?scopes=repo,user,read:org&description=magit-forge
.

It's a bit easier to be able to generate this token with a direct link from the manual.

PS: perhaps this link could be used for automating forge setup?


tarsius:
There's nothing inherently wrong with your suggestion, but I've decided against merging this anyway. (I also don't think this is all that useful. And due to the limitations of texinfo it is ugly.)

PR #512, opened 2022-08-19
customizable heading/body in new pull requests

mrkschneider:
Hello,

first let me say thank you to the maintainers for your excellent work with magit and forge. I've been enjoying both modes immensely.

I wanted to provide a customizable way to adjust the template for new pull requests. As I mostly use Gitlab, I like to use this as a starting point

```
# Merge <source-branch> into <target-branch>

<commit-message>
```

I refactored `forge--prepare-post-buffer` to use two customizable variables setting the functions for adjusting the headline and the body of the PR template. While the default behavior remains as-is, customizing these variables allows to flexibly set the template as e.g. in

```
(defun my-insert-post-buffer-mr-heading (source target)
  (cl-flet ((replace-regex
	     (x)
	     (replace-regexp-in-string "[a-z]+/" "" x)))
    (insert
     (format "# merge %s into %s" (replace-regex source)
	     (replace-regex target))
     "\n\n" )))
```

Admittedly not the prettiest implementation as I don't have a lot of elisp experience, so I'm open for improvements and alternatives if this is at all interesting to you.

### DIFF
--- a/lisp/forge-post.el
+++ b/lisp/forge-post.el
@@ -145,12 +145,10 @@
 (defvar-local forge--pre-post-buffer nil)
 (defvar-local forge--buffer-draft-p nil)
 
-(defun forge-insert-post-buffer-pull-request-heading (source target)
-  (ignore source target)
+(defun forge-insert-post-buffer-pull-request-heading (&optional source target)
   (insert "# "))
 
-(defun forge-insert-post-buffer-pull-request-body (source target)
-  (ignore target)
+(defun forge-insert-post-buffer-pull-request-body (&optional source target)
   (magit-rev-insert-format "%B" source))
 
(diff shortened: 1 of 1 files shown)


tarsius:
(I am actually on a break from responding to requests, but I already looked and this seems easy enough, so let's just do it this one time.)

I don't think that's the right place to do it.  All kinds of "posts" are handled here and it is low level.

I've added the `forge-create-pullreq-hook` instead.  It is more flexible because one can add multiple functions that manipulate the content of the buffer (instead of one function that controls what is being inserted).

You'll need to use the `forge--buffer-base-branch` and `forge--buffer-head-branch` variables to accomplish your goal. I am not documenting the new hook for now; not least because one likely has to use these variables, which were intended to be internal.

PR #517, opened 2022-09-14
Make forge--buffer-draft-p public

### LINKED ISSUE #516: Make forge--buffer-draft-p a configuration option
### Discussed in https://github.com/magit/forge/discussions/509

<div type='discussions-op-text'>

<sup>Originally posted by **DamienCassou** August 18, 2022</sup>
The variable `forge--buffer-draft-p` controls whether the PR to be created will be a draft or not. In some projects, all new PRs should be drafts at first. As a result, it would be helpful to set `forge--buffer-draft-p` as a directory-local variable in these projects. This is already possible but because the variable is private (indicated by the `--` in the name) and because it is not a `defcustom`, this feels awkward.

I suggest creating a `defcustom` for this variable and making it public.</div>

DamienCassou:
Closes #516.

-------

This PR makes `forge--buffer-draft-p` public by transforming the
`defvar` into a `defcustom` and replacing the double dash with a
single one.

There are 3 occurrences of `(local-variable-p 'forge-buffer-draft-p)`
that I left intact (beyond transforming the double dash) as I'm not
sure the exact purpose they serve. I can delete these expressions if
you want me to.

I didn't update the manual as I wasn't sure where to put the
information and the other `defcustom` of `forge-post.el` isn't
discussed there. Please tell me what is the most appropriate place if
you want me to write something and I will happily do it.

### DIFF
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -512,9 +512,9 @@
 
 (cl-defmethod forge--submit-create-pullreq ((_ forge-github-repository) repo)
   (let-alist (forge--topic-parse-buffer)
-    (when (and .yaml (local-variable-p 'forge--buffer-draft-p))
+    (when (and .yaml (local-variable-p 'forge-buffer-draft-p))
       (user-error "Cannot use yaml frontmatter and set `%s' at the same time"
-                  'forge--buffer-draft-p))
+                  'forge-buffer-draft-p))
     (pcase-let* ((`(,base-remote . ,base-branch)
                   (magit-split-branch-name forge--buffer-base-branch))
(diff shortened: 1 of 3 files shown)


tarsius:
> There are 3 occurrences of `(local-variable-p 'forge-buffer-draft-p)`
> that I left intact (beyond transforming the double dash) as I'm not
> sure the exact purpose they serve.

The global value specifies the default state.  The local value is set
when the user decides to change the state of the current pr.

> I didn't update the manual as I wasn't sure where to put the
> information and the other `defcustom` of `forge-post.el` isn't
> discussed there.

Then let's leave it at that for now.

PR #519, opened 2022-09-14
Checkout pull-request without resetting existing branch

### LINKED ISSUE #518: Checkout a PR branch without resetting anything
### Discussed in https://github.com/magit/forge/discussions/511

<div type='discussions-op-text'>

<sup>Originally posted by **DamienCassou** August 19, 2022</sup>
I would like to be able to checkout the already existing local branch corresponding to the PR at point in magit-status. Currently, <key>b f</key> (`M-x forge-checkout-pullreq`) also resets the branch and I don't want that.

Said differently, I would like the same as <key>b b</key> (`M-x magit-checkout`) but with a list of PRs (and the PR at point by default) instead of a list of branches.</div>

DamienCassou:
Closes #518.

-------

This is work-in-progress. I'm opening the PR because I'm not sure I
understand everything and require early feedback. Can you please
advice me?

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -50,6 +50,15 @@ Takes the pull-request as only argument and must return a directory."
   :group 'forge
   :type 'function)
 
+(defcustom forge-checkout-reset-existing-branch
+  nil
+  "srtsrt"
+  :package-version '(forge . "0.4.0")
+  :group 'forge
+  :type '(choice (const :tag "Reset the branch as it exists remotely" t)
+                 (const :tag "Ask the user if the branch should be left intact" ask)
+                 (const :tag "Keep the branch intact" nil)))
+
 ;;; Dispatch
 
 ;;;###autoload (autoload 'forge-dispatch "forge-commands" nil t)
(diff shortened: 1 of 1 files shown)


DamienCassou:
/CC @tarsius I would appreciate early feedback. Thank you.

DamienCassou:
I've seen you have pushed some cleanup fixes, thank you. Do you know if this is the right direction? Is there something I should work on?

DamienCassou:
I've been using that PR for over a month now and am very happy with the result.

lisp/forge-commands.el
```diff
@@ -736,17 +744,22 @@ because the source branch has been deleted"))
...
-                                  (oref pullreq number)))
-           (forge--pullreq-branch-active pullreq))
+     (if (or (not (magit-branch-p branch))
+             (eq forge-checkout-reset-existing-branch t)
+             (and (eq forge-checkout-reset-existing-branch 'ask)
+                  (y-or-n-p (format "Reset %s?" branch))))
          (let ((magit-inhibit-refresh t))
-           (forge-branch-pullreq pullreq))))))
+           (forge-branch-pullreq pullreq))
+       branch))))
```
sellout, line 762: I’m not certain I follow everything here, but this looks like it will ask _twice_ if you select `'ask` (once “Reset %s?”, and then “A branch named "%s" already exists … [r][c][a]” when `forge-branch-pullreq` is called) and if you select `t`, then it will still ask once in `forge-branch-pullreq`, no?

I think this change would be better made a bit deeper, in `forge--pullreq-branch-select`, because it’ll improve both `forge-checkout-pullreq` and `forge-checkout-worktree` (I use the latter mostly).

Then an option could be added to the existing prompt to support checking out the existing branch `[u]se existing "%s" branch` (a result you can’t currently get, and in your change is only possible unilaterally).

And finally, the `defcustom` would have options for every case, like `'reset`, `'create`, `'use`, `'abort`, and `nil` (for “ask”). Leaving `nil` as the default so that the behavior doesn’t change for users who don’t set it.

lisp/forge-commands.el
```diff
@@ -50,6 +50,14 @@ Takes the pull-request as only argument and must return a directory."
   :group 'forge
   :type 'function)
 
+(defcustom forge-checkout-reset-existing-branch nil
```
sellout, line 53: I _think_ with the current state of this branch that the default here should be `t`, as that will preserve the existing behavior. `nil` is “dangerous” in that it gives the user an existing branch without letting them know (and without them choosing that behavior). With the existing code they would only be handed a branch without question if it were newly-created.

(With my suggestions below, `nil` _would_ be the correct default.)

DamienCassou:
Thank you very much for your feedback. This PR is only a draft to get some help from @tarsius as I don't understand everything the code is doing.

I've been using the code for months now and am very happy about it.

tarsius:
I'm sorry this has continuously been falling through the cracks. I'll try to have another look soon.

tarsius:
I've merged something.

The more I looked at the clearer it became that an existing branch should just be checked out and that offering the option to reset would lead down a rabbit whole. More details in the commit message.

PR #522, opened 2022-10-05
Add value to the topic ref section

ccqpein:
The forge side change on [idea](https://github.com/magit/forge/discussions/521#discussioncomment-3800701). 

Give the value to the topic-refs magit section. For opening the magit diff buffer from PR directly.


tarsius:
i plan to support "visit the appropriate issue/pull-request in a buffer" in more places *eventually*, but I intend to do that in one go instead of making it selectively possible in a few random places such as this.

PR #532, opened 2023-01-12
forge--url-regexp: ignore "ssh." subdomains

sebastianpoeplau:
Some instances of GitLab (and possibly other forges) serve their SSH interface at a subdomain (e.g., "ssh.gitlab.some-company"). If the forge is registered in `forge-alist` using the subdomain as `GITHOST`, then `forge-browse` won't work because it derives URLs for the web interface from the `GITHOST`. This commit causes forge to ignore the "ssh." prefix, so that the forge can be registered with its web interface URL in `forge-alist` (e.g., "gitlab.some-company").

Do you think this is the right approach? Or should `forge--format` derive web URLs from the `APIHOST` instead of the `GITHOST`? I've never seen the API being served at another domain than the web interface.


tarsius:
The problem obviously is that, based on the limited data I had, I assumed that the api and web host always use the same subdomain.

The proper way of dealing with that would be to add set the web host separately in `forge-alist`. But users are already confused by the ID and APIHOST usually being the same. And more importantly, the functions for url mangling were written with this assumption in mind. Going down this road we might end up having to duplicate many of them to provide web and api variants, or add arguments to communicate what kind of url we need. `forge--format` also isn't used exclusively for urls.

Here is a (not complete and likely not 100% correct) analysis of what calls what:

```
* forge--url-regexp
  * forge--split-url
    * forge--split-remote-url
      * forge--format (remote slot &optional spec)
    - forge-get-repository
      [url from "git remote get-url <remote>"]
  * forge--url-p
    * forge--forge-remote-p
      - magit-branch-rename
  * forge--url-equal
    - forge--setup-pullreq-remote

* forge--format (obj slot &optional spec)
  - forge-browse-*
  - forge--create-*obj* [for prompt and local filename]
  - forge--edit-*obj* [for prompt and local filename]
  * forge-get-url (obj) [using *-url-format]
    - forge-browse
    - forge-copy-url-at-point-as-kill

- forge-get-repository [url from argument]
  - forge-add-repository [url from prompt]
```

I think it is risky to just drop the `ssh.` part at such a low-level. This might very well affect something that needs it not to be dropped.

Have you only used this commit with a repository that has been added *before* this commit was added, or have you also tried (1) removing an affected repository (2) apply the commit (3) add the repository back (4) use `forge-browse-*`, and (5) use the api? (And the same with a repository that isn't supposed to be affected.)

If that blows up, then we know we cannot do it like this (so please give it a try). If it works, then I think we still should not do it.

A safer approach would be to create variants of `forge--format` and `forge-get-url` for use by `forge-browse*` and such. And by that I don't mean additional methods, but new functions named `forge--format-web-url` and `forge-get-web-url`, to clearly communicate that they can only be used for web urls. This would likely be wrappers around the existing functions, which would just `s/ssh\.//` in the url returned by the wrapped function.

That's hacky of course, but better hacky than sorry.

sebastianpoeplau:
Thanks for the detailed reply @tarsius!

> The problem obviously is that, based on the limited data I had, I assumed that the api and web host always use the same subdomain.

Just to avoid misunderstandings when people read this later, it's the SSH server that is at a different subdomain in my case (i.e., what forge calls the Git host); API and web interface are served at the same domain.

> I think it is risky to just drop the `ssh.` part at such a low-level. This might very well affect something that needs it not to be dropped.

Yes, I see your point.

> Have you only used this commit with a repository that has been added _before_ this commit was added, or have you also tried (1) removing an affected repository (2) apply the commit (3) add the repository back (4) use `forge-browse-*`, and (5) use the api? (And the same with a repository that isn't supposed to be affected.)

I'm using it with a repository that I re-added after the commit. For this case, I confirm that the browsing functionality and API interactions work.

> If that blows up, then we know we cannot do it like this (so please give it a try). If it works, then I think we still should not do it.

Then maybe I should rework my solution before spending time on extensive testing :wink: 

> A safer approach would be to create variants of `forge--format` and `forge-get-url` for use by `forge-browse*` and such. And by that I don't mean additional methods, but new functions named `forge--format-web-url` and `forge-get-web-url`, to clearly communicate that they can only be used for web urls. This would likely be wrappers around the existing functions, which would just `s/ssh\.//` in the url returned by the wrapped function.

I'm not sure that we're talking about the same problem; in my case, the API and the web interface are served at the same subdomain. The problematic part is just SSH access, i.e., the remote URL returned by Git. (Probably the same caveats apply though.)

> That's hacky of course, but better hacky than sorry.

After reading your concerns, I looked at a few alternatives. I can work around the issue using Git's URL rewriting as [suggested in the wiki](https://github.com/magit/forge/wiki/Tips-and-Tricks#using-git-url-rewriting-instead) to prevent forge from ever seeing the SSH subdomain. For example:

```
[url "git@ssh.gitlab.some-company"]
        insteadOf = git@gitlab.some-company
```

The disadvantage is that affected users (1) have to find out about the workaround and (2) use a modified Git remote. In other words, they need to change the clone URL that they likely copy from the web interface, eliminating the `ssh.` portion. URL rewriting then makes sure that Git still uses the subdomain to talk to the server, whereas forge never sees it and happily constructs correct URLs for the API and the web interface.

tarsius:
Sorry for not getting back to you on this again until now.

> Just to avoid misunderstandings when people read this later, it's the SSH server that is at a different subdomain in my case (i.e., what forge calls the Git host); API and web interface are served at the same domain.

Yes, it seems I got confuses about which pair of urls we are talking about.

> After reading your concerns, I looked at a few alternatives. I can work around the issue using Git's URL rewriting as [suggested in the wiki](https://github.com/magit/forge/wiki/Tips-and-Tricks#using-git-url-rewriting-instead) to prevent forge from ever seeing the SSH subdomain.

> The disadvantage is that affected users (1) have to find out about the workaround and (2) use a modified Git remote.

I've decided to require users do that anyway; at least for now. I've added a note about it to `forge-alist`'s docstring; that should take care of (1) at least.

sebastianpoeplau:
> Sorry for not getting back to you on this again until now.

No worries at all!

> > After reading your concerns, I looked at a few alternatives. I can work around the issue using Git's URL rewriting as [suggested in the wiki](https://github.com/magit/forge/wiki/Tips-and-Tricks#using-git-url-rewriting-instead) to prevent forge from ever seeing the SSH subdomain.
> 
> > The disadvantage is that affected users (1) have to find out about the workaround and (2) use a modified Git remote.
> 
> I've decided to require users do that anyway; at least for now. I've added a note about it to `forge-alist`'s docstring; that should take care of (1) at least.

Sounds like a good compromise, thanks!

sin-ack:
Saw the documentation on `forge-alist` and wanted to chime in that I would really like the git URL and web interface being separate as well. The way I have it setup is as follows:

- I have `github-work` and `github` as two different SSH identities, each using a separate IdentityFile in `.ssh/config`.
- Whenever I'm pulling something from GitHub that requires authentication (or I expect to push something) I pull with `git clone ssh://github-work/...`

Because these aren't URLs at all, right now forge tries to open `https://github-work/...` which obviously fails. It would be great if I could tell Forge that all `forge-browse` functions are to use `github.com` instead. Maybe a WEBHOST field?

tarsius:
- The "webhost" is now used, see 50b8dc050f6e7cd1196bb5a3f161161989d585b3.
- When using an ssh alias, it is no longer necessary to add an entry to `forge-alist`. See f537f765baf512b870560361d1cf56dcb1709c2f.

PR #534, opened 2023-01-16
forge-fork: Fix how to get remote name

kamoii:
The manual uses the :remote-name keyword, but the code tries to use 'remote-name
symbol. Fix according to the manual.


tarsius:
Thanks for reporting this. I've fixed the documentation instead; that way nobody's configuration breaks. (Also, the documentation did list `remote-name`. The `:remote-name` typo was only in the example.)

PR #539, opened 2023-02-03
Improve error handling

leotaku:
This PR aims to improve on the fixes introduced in [8ab946](https://github.com/magit/forge/commit/8ab946a5eb1e8de8e54e02f21e65dbd1b97ceff3).

The main idea is that most entry points to Forge call `forge-get-repository` early on, so error handling can be improved by adding a check for whether the database can be accessed to that function. Personally, I think this is a quite nice solution, but others might disagree. Also, I am not completely sure I have understood the purpose of the `demand` argument entirely, so my implementation might be subtly broken.

The new error handling also allows us to remove the `forge--sqlite-available-p` checks. This is desirable because they can cause Forge to be deactivated when a SQLite executable isn't present, even with backends like `sqlite-builtin` that don't require a SQLite executable.

### DIFF
--- a/lisp/forge-db.el
+++ b/lisp/forge-db.el
@@ -90,10 +90,6 @@ by the `sqlite3' package.  You need to install thef
        ((object-class :initform 'forge-repository))))))
 
 (defconst forge--db-version 9)
-(defconst forge--sqlite-available-p
-  (with-demoted-errors "Forge initialization: %S"
-    (emacsql-sqlite-ensure-binary)
-    t))
 
 (defvar forge--db-connection nil
   "The EmacSQL database connection.")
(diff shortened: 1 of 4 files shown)


tarsius:
Sorry for not replying earlier.

You are of course right, now that we support three different backends, this assert is not appropriate anymore.

We have to do something similar for all three backends now, and that is quite complicated. I'll have to implement that myself and will try to get it done over the next few days.

leotaku:
No need to apologize! 

I'm sure you know the codebase much better than I do, so you will be able to implement a more reliable solution. Feel free to close this PR if you want.

tarsius:
I am doing this as part of a set of changes that span across six repositories. :scream: 

Here's the first published result of that effort: https://github.com/melpa/melpa/commit/4872ef038dbbf67008bfa7951574ee372d6ff68d.

Issue #158, opened 2019-06-03
Add customization to not enabling bug-reference modes everywhere.


thomasf:
I do want it in the forge-topic modes when reading issue comments but it's a bit too distracting for me and overrides other fortification modes.   Maybe a defcustom to for limiting where it inserts itself would be nice?

I quickly solved it for myself with an advice, not really sure which modes I want this in.. 

```el
(defun my-forge-bug-reference-setup-condition (orig-fun &rest args)
  (when (or (eq major-mode 'forge-topic-mode)
           (eq major-mode 'magit-staus-mode)
           (eq major-mode 'forge-issue-list-mode)
           (eq major-mode 'magit-log-mode))
    (apply orig-fun args)))
(advice-add 'forge-bug-reference-setup :around #'my-forge-bug-reference-setup-condition)
```

tarsius:
You can use the new option `forge-bug-reference-hooks` for this now.

PR #543, opened 2023-02-20
Handle saving max topics

elken:
Allows configuration of the forge.max-topics git option for large Github repositories

Depends on https://github.com/magit/ghub/pull/158

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -96,7 +96,9 @@ Takes the pull-request as only argument and must return a directory."
     ("a  " "add repository to database" forge-add-repository)
     ("r  " "forge.remote"  forge-forge.remote)
     ("t t" forge-toggle-display-in-status-buffer)
-    ("t c" forge-toggle-closed-visibility)]])
+    ("t c" forge-toggle-closed-visibility)
+    ("t l" "GQL entity limit" forge-forge.graphqlItemLimit
+     :if (lambda () (forge-github-repository-p (forge-get-repository nil)))) ]])
 
 ;;; Pull
 
@@ -933,6 +935,12 @@ This only affect the current status buffer."
       (setq forge-topic-list-limit (cons forge-topic-list-limit 5))
     (setcdr forge-topic-list-limit (* -1 (cdr forge-topic-list-limit))))
   (magit-refresh))
+
+(transient-define-infix forge-forge.graphqlItemLimit ()
+  "Change the maximum number of GraphQL entities to pull at once."
+  :class 'magit--git-variable
+  :variable "forge.graphqlItemLimit"
+  :reader #'read-string)
 
 ;;;###autoload
 (defun forge-add-pullreq-refspec ()
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -51,7 +51,10 @@
                            &optional callback)
   (let ((buf (current-buffer))
         (dir default-directory)
-        (selective-p (oref repo selective-p)))
+        (selective-p (oref repo selective-p))
+        (ghub-graphql-items-per-request
+         (string-to-number
+          (or (magit-get "forge.graphqlItemLimit") "100"))))
     (ghub-fetch-repository
      (oref repo owner)
      (oref repo name)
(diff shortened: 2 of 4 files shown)


lisp/forge-commands.el
```diff
@@ -893,6 +895,13 @@ is added anyway.  Currently this only supports Github and Gitlab."
 
 ;;; Misc
 
+(transient-define-infix forge-forge.max-topics ()
+  "Change the maximum number of topics to pull at once."
+  :class 'magit--git-variable
+  :variable "forge.max-topics"
+  :reader (lambda (&rest args)
+            (apply #'read-string args)))
```
tarsius, line 903: Is this really necessary? I think this should fall through to using `read-string` if no reader is given explicitly.

lisp/forge-commands.el
```diff
@@ -95,6 +95,8 @@ Takes the pull-request as only argument and must return a directory."
   [["Configure"
     ("a  " "add repository to database" forge-add-repository)
     ("r  " "forge.remote"  forge-forge.remote)
+    ("t l" "max topics to pull" forge-forge.max-topics
+     :if (lambda () (forge-github-repository-p (forge-get-repository nil))))
```
tarsius, line 99: Not sure if that is possible, but have you tried moving that into the infix definition?

elken: I was just mirroring an example of one further up :smile:

lisp/forge-commands.el
elken: You're spot on there :sweat_smile: Thanks

elken:
Just dawned on me I haven't added any docs for this, should I throw something in the manual?

tarsius:
I've just looked at this too, and it looked wrong, but I decided to not deal with it now. But since you are around right now, we can deal with it now.

lisp/forge-commands.el
tarsius: Yeah, no need to change that.

But please it last and on "level 7".

lisp/forge-github.el
```diff
@@ -51,7 +51,8 @@
                            &optional callback)
   (let ((buf (current-buffer))
         (dir default-directory)
-        (selective-p (oref repo selective-p)))
+        (selective-p (oref repo selective-p))
+        (ghub-graphql-items-per-request (string-to-number (or (magit-get "forge.max-topics") "100"))))
```
tarsius, line 55: That line is too long.

elken: Ah you're right.

lisp/forge-commands.el
```diff
@@ -893,6 +895,12 @@ is added anyway.  Currently this only supports Github and Gitlab."
 
 ;;; Misc
 
+(transient-define-infix forge-forge.max-topics ()
+  "Change the maximum number of topics to pull at once."
+  :class 'magit--git-variable
+  :variable "forge.max-topics"
```
tarsius, line 901: We usually do what Git does and use camelCase. `forge.graphqlItemLimit` and `forge-forge.graphqlItemLimit` sound about right.

Also, since this is less important than other infixes, place it this after the definitions of those, probably right before the database utilities at the end of this section.

elken: Sure, I wasn't sure which convention to prefer and I'm sure I found a dashed one; not a problem to change :)

elken:
I'm adding a note to "Github caveats" in the manual too

elken:
> Please also change the commit message to be more in line with this repositories conventions.

I see both semantic commits and regular are used, which is preferred?

tarsius:
It depends on what kind of change it is. Here we are just adding a fairly simple thing, so just say that. `forge...: New command`. Of course feel free to say more in the rest of the message; including the why.

tarsius:
Please also rebase. (And I am currently looking into warning/error about compat-alist-get`.)

elken:
> Please also rebase. (And I am currently looking into warning/error about compat-alist-get`.)

I have that queued up :)

elken:
Is there a reason you force pushed this to a wrong commit? :P

tarsius:
No, that was an accident, sorry. Please push your latest version again.

elken:
> No, that was an accident, sorry. Please push your latest version again.

No harm no foul :)

PR #577, opened 2023-05-05
manual: fix typo

dkm:
Fix typo in probable copy/paste.


dkm:
Tried to generate forge.texi, but it ended up changing everything in the file... Probably a versions mismatch, so left it out (as instructed in the contrib text).

tarsius:
Thanks!

PR #610, opened 2023-12-14
forge: added support for gitea

mobid:
I'm using self hosted gitea on regular basis, so here is the result.

From day to day use I found two bugs:
1. `forge-add-repository` doesn't pull topics at first time (same behaviour has github)
2. `forge-pull-notifications` most times throw error:
   > error in process filter: json-read: Unrecognized keyword: "f51"
   ^ Probably broken read of chunked http response?

Tested:
- [X] Pull request
  - [X] Create
  - [X] Assign
  - [X] Close/Open
  - [ ] Drafts (no drafts, only WIP prefix)
  - [X] Milestones
  - [X] Labels
  - [X] Assignees
  - [X] Review-Requests
- [X] Issues
  - [X] Create
  - [X] Assign
  - [X] Close/Open
  - [X] Milestones
  - [X] Labels
  - [ ] Marks (there is no such thing in gitea?)
  - [X] Assignees
- [X] Commenting (comment, edit, delete)
- [X] Merge
- [ ] Notifications - need another check
- [ ] Templates - need another check

Any suggestions/tips are welcome.

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -215,7 +215,8 @@ If pulling is too slow, then also consider setting the Git variable
   (interactive)
   (if-let ((repo (forge-get-repository 'maybe)))
       (let ((class (eieio-object-class repo)))
-        (if (eq class 'forge-github-repository)
+        (if (or (eq class 'forge-github-repository)
+                (eq class 'forge-gitea-repository))
             (forge--pull-notifications class (oref repo githost))
           (user-error "Fetching notifications not supported for forge %S"
                       (oref repo forge))))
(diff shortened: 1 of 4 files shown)


tarsius:
This has been suggested a few times before.

There also was a pull-request that added notification support (#431).  In response to that I started working on that myself.  The early results can be seen in the `[dump] support gitea` commit on the `idea` branch.

But I abandoned that work.  I now think that it was a mistake to support more than one forge initially; it slows down development considerably.  The Gitlab implementation, which most other implementations would be based on, is lacking.  One problem is callback hell.  Until I have improved the Gitlab implementation, I don't want to add other implementations that are based on that.

So I'm afraid you will have to maintain that as a separate package and you will also end up having to deal with breaking changes.  I'm sorry I cannot do better than that at this time.  Forge is a big time sink and I am prioritizing paying of technical dept over supporting more forges (which would mean increasing the dept).

mobid:
Ok. I will pack it separately. Thank you.

PR #613, opened 2023-12-27
forge-browse-remote: auto-accept single candidate

vermiculus:
Using `N b r` in a repository with only a single 'origin' remote would still prompt the user to select a remote to browse.

Given this is a non-destructive action (and not something like `magit-remote-prune', for example), there is no need to prompt when there's only one option.


tarsius:
Thanks!

PR #628, opened 2024-02-14
Optionally show notification counts

necaris:
Add a new custom option for whether to always try and show counts, and
if it's set add a count of matching notifications to the heading.

### DIFF
--- a/lisp/forge-notify.el
+++ b/lisp/forge-notify.el
@@ -45,6 +45,12 @@
   :group 'forge
   :type (if (>= emacs-major-version 28) 'natnum 'number))
 
+(defcustom forge-notifications-always-show-counts t
+  "Whether to try to always show a count of items."
+  :package-version '(forge . "0.4.0")
+  :group 'forge
+  :type 'boolean)
+
 ;;; Class
 
 (defclass forge-notification (forge-object)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks for the contribution!

Unfortunately, from a [or this] maintainer's perspective, adding an option for such a minor detail is a worse choice than to either never or always showing these counts. It adds only minimal maintenance overhead, but these things accumulate.

Also, whatever we would pick as the default, it is likely that there would be users who would prefer the other choice, but will never learn about the existence of the option, because they too consider it a minor detail and would not expect an option to exist.

Of course I might be wrong about that, so maybe I should merge the addition of an option anyway. We could call that the *maintainer's second-order choice dilemma*. :grin:

Anyway, I am currently dealing with much larger issues, and will delay the decision about this until I am ready to return to and focus on such details.

necaris:
Thank you for the gracious response! Very much appreciate the work you're doing on the whole `forge` ecosystem here.

> Also, whatever we would pick as the default, it is likely that there would be users who would prefer the other choice, but will never learn about the existence of the option, because they too consider it a minor detail and would

For what it's worth, I'd personally prefer to always show the counts (without an option) but didn't want to clutter the display if that's not desired. I can also think of other ways to customize the headings, so people who really care (like me 😉) can do what they wish.

> Anyway, I am currently dealing with much larger issues, and will delay the decision about this until I am ready to return to and focus on such details.

Of course, thank you for considering this change. Best of luck resolving the other issues.

tarsius:
Since 17f5fa39db99167a2dd9f1f0e53868c14fd3f25f you can add an :around advice for the new `forge-notifications-buffer-desc` function, to append the count.

necaris:
thank you!

tarsius:
Thanks for you patience. :grinning:

PR #635, opened 2024-02-19
transient-define-suffix: Fix autoload

benthamite:
Commit magit/forge@0a656ed replaced a function with a transient suffix, but it left the associated autoload unchanged. As a consequence, a warning is generated at startup ("void-function transient-define-suffix"). This patch defines the autoload correctly.


tarsius:
Thanks!

PR #638, opened 2024-02-20
Allow browsing the remote from `forge-dispatch` even when repo is not in the database

benthamite:
This pull request allows the user to browse the remote from the Forge dispatcher even when the repository has not been added to the database. Browsing a remote is a very common operation and it seems like it should be possible to perform it without having to add a repo (or define a custom dispatcher or key binding).

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -102,6 +102,7 @@ Takes the pull-request as only argument and must return a directory."
                      "Not inside a Git repository"))
     :if-not forge--get-full-repository
     ("a" "add repository to database" forge-add-repository)
+    ("b" "browse remote"              forge-browse-remote)
     ("f" "fetch notifications"        forge-pull-notifications)
     ("l" "list notifications"         forge-list-notifications)]])
 


benthamite:
(Another, perhaps more consistent, approach would be to create a new dispatcher for all the commands currently in `forge-dispatch` that can be invoked even when the repository is not in the database.)

tarsius:
> Allow browsing the remote even when repo is not in the database

Well, it's already allowed, the command just isn't bound where you would expect it.

> Another, perhaps more consistent, approach would be to [...]

I am currently refining synchronizing the status of topics between Github and Forge. (Which is hard because that part of the API is abysmal.)

After that I'll return my attention to key bindings and related issues. I'll delay thinking about this particular detail, until then.

benthamite:
> Well, it's already allowed, the command just isn't bound where you would expect it.

Slightly edited the title to clarify.

tarsius:
`forge-dispatch` now offers some more commands for untracked repositories.

PR #703, opened 2024-09-17
Respecting `auto-revert-remote-files` in `forge-bug-reference-setup`

bramadams:
Fixes a similar issue as https://github.com/magit/magit/issues/5222, where, when using tramp, forge would do substantial unnecessary work inside `forge-bug-reference-setup`, slowing down the connection or even hanging.

To fix this issue, changes similar to the original magit commit (see https://github.com/magit/magit/commit/61c051ea1cda5fe6c9404cb5ae228088d2e254f0) could easily be repeated, basically shortening out the `unless` within `forge-bug-reference-setup`.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -1735,6 +1735,9 @@ enable `bug-reference-mode' or `bug-reference-prog-mode' and
 modify `bug-reference-bug-regexp' if appropriate."
   (unless (or bug-reference-url-format
               (not (forge-db t))
+              (and (not auto-revert-remote-files) ; see magit's #5422
+                   buffer-file-name
+                   (file-remote-p buffer-file-name))
               ;; TODO Allow use in these modes again.
               (derived-mode-p 'forge-topics-mode 'forge-notifications-mode))
     (magit--with-safe-default-directory nil


tarsius:
`auto-revert-remote-files` purpose is to control an unrelated feature, and should not be repurposed here.

> when using tramp, forge would do substantial unnecessary work

What part is expensive?  Did you do some benchmarking?

I don't think it is "unnecessary" in the same way as was the case for magit/magit#5222.  This does cause bug-references to be by "linkified", doesn't it?  Doing that *might* be too expensive to be *worth it*, but "unnecessary" doesn't seem like the appropriate term, unless I am missing something.

Can you say more about the hanging? Why does that happen?  Does that actually happen, or is the process just so slow, that you always abort before it completes?

bramadams:
Thanks for checking this PR!

For a given remote file, it basically hangs for at least a minute, before I manually abort. Then, if I successively open the same file, it opens. When trying a different file, the second try would work again. Neither of these files are inside a git repository, just a regular folder where I have read/write access. When invoking tramp to access those files, I'm also not inside any local git repository.

Earlier today, I made a screenshot of the stack trace after sending a SIGUSR2 to emacs while it was hanging, see attached. That's where I noticed `forge-bug-reference-setup`. Also, if I toggle the new variable `forge-bug-reference-remote-files` to `nil`, the hanging disappears.

stack_trace

bramadams:
@tarsius Just a heads-up that I've rebased the PR up to forge 0.4.5. I can confirm that the PR has been working perfectly for the past 3 months (no more hangs), both on 29.x and the latest 30.x pretest. Is it possible to merge this?

tarsius:
Happy new year!

bramadams:
Thanks a lot, and happy holidays to you too!

PR #713, opened 2024-10-14
Fix token generation link for GitLab

shutosheep:


shutosheep:
Sorry, I messed up a little bit. I think title and description are written properly now.

Hello. I noticed that token generation link for GitLab in the manual points to a broken link. This PR updates the broken link.

PR #744, opened 2025-01-21
forge--msg: Make it more resilient to buffer kills

LemonBreezes:
Previously, if you killed the `magit-status` buffer while forge was
fetching the repository information, and ran `magit-status` again, the
forge status would be missing from the modeline. With this PR, we use
the buffer name as a backup to identify the status buffer in case it was
killed.

The reason why I want this change is that by default, Doom Emacs kills the
`magit-status` buffer when you quit magit "if the buried/killed magit buffer was
the last magit buffer open for this repo" (https://github.com/doomemacs/doomemacs/blob/master/modules/tools/magit/autoload.el#L125).


tarsius:
Closes #633.

LemonBreezes:
Force-pushed because I realized I did not use `buffer-live-p` for `buf`.

PR #767, opened 2025-04-12
Change the suffix for magit-dispatch to avoid an error

jmcarcell:
I have updated to the last version of magit and forge and I'm finding the following error:
```
Error (use-package): forge/:catch: Cannot insert ("N" "Forge" forge-dispatch) into magit-dispatch; o not found
```
There is a related issue: https://github.com/magit/forge/issues/388
and this solution is the same as highlighted in one of the comments:
https://github.com/magit/forge/issues/388#issuecomment-899708964
After this change the error doesn't appear anymore.

### DIFF
--- a/lisp/forge.el
+++ b/lisp/forge.el
@@ -124,7 +124,7 @@ is loaded, then `magit-mode-map' ends up being modified anyway.")
   (keymap-set magit-commit-section-map "C-c C-v" #'forge-visit-topic)
   (keymap-set magit-branch-section-map "C-c C-v" #'forge-visit-topic)
 
-  (transient-insert-suffix 'magit-dispatch "o"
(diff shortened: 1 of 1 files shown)


tarsius:
You forgot to mention that this happens because you are using `evil-collection`.

The suffix in `magit-dispatch` are sorted alphabetically by key. I do not know to what extend `evil-collection` messes that up, but it does not make sense to insert this command out-of-order for everyone, just for the benefit of `evil-collection` users.

Failure to insert a suffix did not result in an error until recently. I have now gone back to that behavior by default, see https://github.com/magit/transient/discussions/374.

`evil-collection` should be adjusted to add `forge-dispatch` to `magit-dispatch` in a location and using a key that is appropriate when using `evil`.

PR #768, opened 2025-04-12
Offer blank option in addition to templates

artenator:
## Summary
Allow for choosing "Blank" as an alternative to templates when they're available.

This PR makes it so that if there are templates available for Issues or PR topics, then an additional "blank" option is offered.

## Reasoning
Sometimes one wants to create an issue or pull request without using a template. This also aligns with how github presents it in their web app:
image

## Additional notes
I saw that `'forge-discussion` is a handled topic in the case statement for the completing-read below, but it was not permitted to get into that branch because of the equality check. I assumed this meant that it's not something that's supported right now and may be later on and modeled it as such, but let me know if I was incorrect with that assumption.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -1933,8 +1933,11 @@ alist, containing just `text' and `position'.")
 
 (cl-defmethod forge--topic-template ((repo forge-repository)
                                      (class (subclass forge-topic)))
-  (let ((choices (and (not (eq class 'forge-discussion))
-                      (forge--topic-templates-data repo class))))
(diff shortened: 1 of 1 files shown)


tarsius:
The maintainer [can force](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser) users to use a template when using the web-interface and Forge should respect that.

I've implemented that on the `yaml` branch but because I want to make additional changes in this area, I'm not merging this just yet.

tarsius:
The eagle has landed.

Issue #171, opened 2019-07-14
Base and head repos/refs are not displayed for pull requests


darth10:
This is something fairly basic shown in GitHub.
Since it's available on the API, it should be easy to surface.

![image](https://user-images.githubusercontent.com/1171466/61182576-ae820e00-a689-11e9-80ed-eccf8f783a4f.png)

I've done some work on this in #168.

tarsius:
I have seen your ping but was busy the last two days. There is no need for this duplicate.

Issue #170, opened 2019-07-11
error in process sentinel: forge--color-brightness: Color doesn’t have #NNNNNN format


davep:
This seems to have started coinciding with forge updating to `forge-20190708.2203` for me. If I `forge-pull` I get the following error message showing:

```
forge--color-brightness: Color doesn’t have #NNNNNN format
```

I modified `forge--color-brightness` to report the colour it wasn't happy with when throwing the `error` and, in this case, it's a colour value of "#666".

I'm running against a locally-hosted GitLab and there is such a label colour:

![image](https://user-images.githubusercontent.com/28237/61038156-69798580-a3c4-11e9-807b-f69fbf0a26a3.png)

tarsius:
What is the six digit representation of `#666`? `#606060`?

davep:
I believe so, yes. It's one of the valid formats listed here: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#RGB_colors

tarsius:
According to that page it is the same as `#666666`.

davep:
Yes, sorry, misread the above. The latter is correct.

tarsius:
That should do it.

davep:
Thanks, I'll give it a test very soon.

```elisp
(error "Color does not have #RRGGBB or #RGB format")
```

It might be worth noting, from the above link, that something like `#6666` is also valid (as is `#66666666`). For example:

![image](https://user-images.githubusercontent.com/28237/61041718-b82a1e00-a3ca-11e9-95af-a63ae17fd091.png)
![image](https://user-images.githubusercontent.com/28237/61041759-cd9f4800-a3ca-11e9-9deb-548d59be6bab.png)
![image](https://user-images.githubusercontent.com/28237/61041794-dc85fa80-a3ca-11e9-859d-d20119ca775d.png)

Etc...

tarsius:
Emacs has some code to determine whether to use the dark or light face definition and given that this is getting more complicated than I originally thought it would probably be better to locate and reuse that than to re-implement it. Maybe you could help me locate it?

davep:
I don't know of the feature you mention, but if I get some time I'd be happy to have a dig around (which will likely mean trying to better understand what forge is doing here too). Might be a few days I'm afraid as I've got quite a lot on at the moment.

davep:
Would it be helpful for me to open this as a new issue (that way it should serve as a reminder to me that I need to take a look)?

tarsius:
[...]

PR #770, opened 2025-04-16
WIP: Add forgejo support

pinoaffe:
Add support for forgejo

Note: Nearly everything worked at some point (apart from the FIXME's), but I've since made changes and have yet to retest everything, so this may currently be broken

This depends on https://github.com/magit/ghub/pull/171

### DIFF
--- a/lisp/forge-forgejo.el
+++ b/lisp/forge-forgejo.el
@@ -306,26 +306,23 @@
                (pullreqs "PRs")))
         i cnt)
     (lambda (cb topics &optional val)
-      (unless cnt
-        (setq i 0
-              cnt (length topics)))
-      (cl-incf i)
-      (if topics
-          (progn
-            (forge--msg nil nil nil "Pulling %s %s/%s" typ i cnt)
-            (forge--fetch-issue-posts
-             repo topics
-             (lambda (cbb)
-               (if (or (eq field 'issues)
-                       (not cbb))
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks, but I do not intend to merge this.

You could maintain this in a separate repository and take full responsibility for it, while we wait for me to implement support in forge directly.

Forge has a lot of technical depth. For months now I have been working on slowly paying it off. Next month I plan to turn my focus on paying of the technical depth related to supporting another forge in addition to github (i.e., gitlab). Once that is payed off it will be possible to add support for additional forge, *without going deeper into depth*.

This pr is based on someone else's pr, #610, which wanted to add gitea support. The reasons I gave there for not merging that, also apply here. By the ways, after #610 I actually did work on gitea support myself a bit (because of the reason mentioned at https://github.com/magit/ghub/pull/171#issuecomment-2812162008), but it quickly became clear that indeed it would be better to first improve `forge-gitlab.el`, before copying that to `forge-gitea.el` and adjusting to actually support gitea.

That being said, if you actually do end up maintaining a `forge-forgejo.el`, I think you would be better of basing that off a contemporary `forge-gitlab.el` instead of the old `forge-gitea.el` you picked.

As mentioned at https://github.com/magit/ghub/issues/170#issuecomment-2780638640, I do actually plan to add support for Forgejo, but this is still a few months off. It most likely will happen this year though.

PR #773, opened 2025-05-01
forge--topic-parse-template: Fix parsing when there is no metadata

DamienCassou:
The returned template alist should associate "text" with a string, not
with a function.


tarsius:
Thanks.

PR #774, opened 2025-05-01
forge--topic-template-files: Fix to work in more cases

DamienCassou:
The previous implementation was only working when the filename
containing the PR template was exactly
".github/PULL_REQUEST_TEMPLATE": no extension, all uppercase and
within the .github/ directory. This commis fixes that so the ".md"
extension is allowed, case doesn't matter and the file can also be at
the root of the repo or within docs/..

### DIFF
--- a/lisp/forge-github.el
+++ b/lisp/forge-github.el
@@ -1047,9 +1047,29 @@
 
 (cl-defmethod forge--topic-template-files ((repo forge-github-repository)
                                            (_ (subclass forge-pullreq)))
-  ;; The web interface does not support multiple pull-request templates,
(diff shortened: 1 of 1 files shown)


tarsius:
I'm in the middle of refactoring template handling. I'll also be gone for a few days, and will look into this when I get back.

DamienCassou:
Should I close this?

tarsius:
It's fine to leave this open as a todo item.

tarsius:
I've pushed my version.

DamienCassou:
Thank you!

PR #795, opened 2025-07-01
forge--fork-repository: Fix several issues

mjoerg:
This addresses some issues encountered in `forge--fork-repository`:
+ The poll using `ghub-wait` was always targeting the forge's default host, and therefore timing out with an error message for other hosts.
+ The comparison of the fork destination would always be done against the username set for github's default host.
+ The Gitlab endpoints were not correctly escaped in case the owner or fork destination were a subgroup separated by a slash.


tarsius:
Thanks!

PR #802, opened 2025-07-22
Support GitLab issue templates without YAML frontmatter

braineo:
GitLab currently does not support YAML frontmatter in issue templates (see [gitlab#326905](https://gitlab.com/gitlab-org/gitlab/-/issues/326905)), but Magit Forge's `forge-create-issue` expects templates to have YAML headers with `name` and `about` fields. 

When using templates without YAML frontmatter, the issue selection interface shows unhelpful entries like for the repo I work with:

```
nil
nil - nil
```

This PR improves the template parsing logic to handle templates without YAML frontmatter and improves the display of incomplete template metadata.

### Cases handled:

| Template condition               | Before                             | After                              |
|----------------------------------|------------------------------------|------------------------------------|
| No YAML header                   | `nil`                              | `bug` (from filename)              |
| YAML with missing `name`/`about` | `nil - nil`                        | `feat` (from filename)             |
| YAML with only `name`            | `Bug Report - nil`                 | `Bug Report`                       |
| Complete YAML                    | `Bug Report - Create a bug report` | `Bug Report - Create a bug report` |

Since the change does not affect public forge API, do let me know if i should update the documentation.

### DIFF
--- a/lisp/forge-topic.el
+++ b/lisp/forge-topic.el
@@ -1960,7 +1960,7 @@ When point is on the answer, then unmark it and mark no other."
               (magit-git-insert "cat-file" "-p" file)
               (if (equal (file-name-nondirectory file) "config.yml")
                   (forge--topic-parse-template-config)
-                (list (forge--topic-parse-template)))))
+                (list (forge--topic-parse-template (file-name-base file))))))
           (forge--topic-template-files repo class)))
 
 (cl-defgeneric forge--topic-template-files (repo class))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

I've done it slightly differently.

braineo:
> > I'm totally fine with going back-and-forth to get things right (commit messages and the code) rather than you having to rework them.
> 
> 
> 
> Just doing it myself tends to be a lot less work than coaching someone else, so that's what I tend to do in these situations. I also don't usually know whether someone just wants one issue fixed (and though that would be done quicker if they provide a fix themself), or whether they want to keep contributing long term. Now that you have told me that it's the latter in your case, I plan to review your future pull-requests, instead of just using them "for inspiration".

Thank you for your kind words. That said I seldomly find bugs and the software is very well maintained.

PR #809, opened 2025-09-16
docs: github.com > codeberg.com under "tell forge about the instance"

slackline:
Fixes a minor tpyo in the text preceding the example entry for `codeberg.org` entry under `forge-alist` in the *Tell Forge about the Instance* section.

> Edit only "forge.org".  To update "forge.texi" run "make texi".

`forge.texi` is included, it was generated via `make texi` (based on looking at earlier PR #577 which also fixed a tpyo in the manual and included both `.org` and `.texi`). Hope that is correct.

### DIFF
--- a/docs/forge.org
+++ b/docs/forge.org
@@ -479,7 +479,7 @@ at least enables the use of commands such as ~forge-browse~.
 A few hosts, which use partially supported forge types, are available
 out-of-the-box, because they have an entry in the default value of
 option ~forge-alist~ (also see its docstring).  For example, the entry
-for https://github.com in that variable looks like this:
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

> `forge.texi` is included, it was generated via `make texi` (based on
> looking at earlier PR #577 which also fixed a tpyo in the manual and
> included both `.org` and `.texi`). Hope that is correct.

Sure, but I had to undo the unrelated whitespace fixes because the Org version I have installed wants to insert those, so if I didn't do it now, they would be added back the next time the manual is edited.

svjson:
"tpyo". I see what you did there. :)

Issue #58, opened 2019-01-03
Show more reviews (and similar "events"), alongside topic posts


mogorman:
We use the github feature for marking a pr reviewed. the posts it generates in the topic don't seem to show up for me. Other posts entered on the pr do seem to show up so not sure where the problem might be.

tarsius:
Github doesn't treat this events as "posts". But this that information is useful and eventually Forge will fetch and display it.

vermiculus:
Design-wise, would you consider them parts, a post subclass, or a different type entirely?

mogorman:
cool @tarsius changed topic. thanks for info. magit is amazing as im sure  you already know.

tarsius:
I don't know yet.

We could also treat posts as a subtype of a new event type.  That's what Gitlab does.

So with Gitlab we get those non-post events for free (especially after thatb019dfa6c9d3ff084b209837fce638bf4f117d5b), but there are other problems.  As mentioned in that commit, events have no "subtype" field (at least I didn't see something like that when I last checked).  And Gitlab additionally supports "threads" which are secondary discussions attached to a topic that are separate from the "normal posts" and "non-post events".  I haven't looked into this any further, but it looked messy at first sight.

thomas-riccardi:
Github CI integration result is another type of information about a topic that could usefully be displayed in forge.

tarsius:
My excitement about Github (and Gitlab) has reached a new low-point -- the stochastic parrots pushed me over the edge. Of course I/we won't be able to avoid Github altogether, because many projects use Github and will continue to do so. So I will continue to maintain Forge, even adding a new feature here and there, but I plan to move my own repositories elsewhere, and significantly reduce the time I put into new Forge features. 

Maybe I'll move to Codeberg. Or I might go one step further and embrace the "distributed" in "Git is a fast, scalable, distributed revision control system". We probably should have done that all along. I most likely won't drop back to an email-based workflow -- that never satisfied, not even close.

Investing all the exiting alternatives, settling on one+ and implementing tooling will be a lot of work, but at least I am excited about it, unlike when I work on tooling that helps people stay on Github.

So while I always planned to implement the requested feature "eventually", I have now arrived at the point where I finally allow myself to say "no" to features that are not absolutely essential.

PR #819, opened 2025-10-07
forge--update-pullreq: Fetch and store review comments

braineo:
Fetch and store review comments alongside with normal comments.

Before this change, the review comments are not visible in pullreq view like following

image


braineo:
To demonstrate the idea I'll create normal comments vs review comments in the PR. This should be a normal comment that is display in the pullreq view

braineo:
@tarsius, actually do you think this pr is related to #58?

PR #829, opened 2026-01-11
Improve default branch detection for pull requests

DamianB-BitFlipper:
Currently only `master` is used as a last resort fallback. A lot of repos nowadays use `main` so I added `main` as well before `master`.


tarsius:
Thanks, but honestly, this seems a bit excessive -- a lot has to be misconfigured for this to even become relevant, and if we get there it just means that the default might not be right, not that the appropriate branch cannot be selected.  If you had only asked for this feature, without providing an implementation, it would probably have ended up on the backlog.

But if we do this, then we should take it a step further and consider all branches listed in `magit-main-branch-names`. Which I have done now.

PR #868, opened 2026-08-10
forge-visit-topic-from-url: Visit any topic from configured forges, regardless of if it is in the database or the repo tracked

lucasec:
This is the first part of a broader workflow I have been working on: a `browse-url` handler that allows intercepting links to `github.com` and other forges from gnus emails, Org-mode buffers (when the original link was captured without the use of `orgit-forge`), IRC chats, etc. and redirecting those to forge instead of the default browser.

This PR proposes an extension to `forge-visit-topic-from-url` that allows visiting any topic URL from a configured forge in `forge-alist`, regardless of it is already pulled. Essentially, it automates the workflow of trying to use the existing `forge-visit-topic-from-url`, encountering an error because the topic is not pulled (or repo not tracked), and then tracking down a subtitle magit buffer to do one or both of the prior. The proposal:

1. Does not change Forge's existing model of pulling topics into a local database
2. If the repository is already tracked and the topic exists in the database, `forge-visit-topic-from-url` behaves as before
3. If the topic does not already exist, a topic buffer is created in a new "loading" (or "stub") state
    - If the repository is not yet tracked, the `forge-add-repository` transient is automatically opened so the user can choose how to add the repository to the database
    - If the repository is tracked but the topic does not exist in the database, triggers an automatic `forge-pull-topic` to fetch the topic, then (if the topic is a pull request and the repository is associated with a local worktree) fetches the relevant git ref.
    - When loading is complete, the topic buffer automatically refreshes to display the topic.

The implementation of the "loading" topic buffer uses a simple state machine that advances by the existing automatic refresh of the buffer that occurs when the various magit/forge fetch commands complete.

In a second (standalone) commit, I made a few minor tweaks to the `forge-add-repository` UI to hide certain elements in the transient that are irrelevant to the flow of adding repositories by URL.

## Potential areas for improvement

In general, I think the implementation came out fairly clean, which hopefully means it can integrate without disrupting any of your longer-term plans for the package.

There are a few areas that could be improved over time or with a few broader architectural changes (I left most of these marked with TODO comments):
- Forge doesn't really have a good progress/error callback model, so I broke abstraction in one case (`forge-add-repository` sets a buffer-local status variable) and in the other case just had the state machine guess, in order to display some indication to the user of whether the topic is still loading or has stopped due to some error.
- GitLab doesn't support fetching a single topic at the moment, so I just stubbed out this method to the full topic fetch.
- `forge--fetch-pullreq-ref` seems like a useful workflow to generalize, maybe it should always be hooked into `forge-fetch-this-topic` so pressing G in an existing topic buffer both pulls the forge data and any updated git ref (which would be needed to display any new commits) in one go.

## Screenshots

Visiting a topic in a repo that is not yet tracked:
Screenshot 2026-08-09 at 10 33 22 PM

Fetching a topic not yet in the database:
Screenshot 2026-08-09 at 10 34 59 PM

### DIFF
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -1278,8 +1278,8 @@ upstream remote."
         (propertize (forge--scope 'url) 'face 'bold)))
      :format "%d")]
 
-   ;; Nothing to tracked.
-   [:if-not (##forge--scope 'topdir)
+   ;; Nothing to track.
+   [:if-not (##or (forge--scope 'topdir) (forge--scope 'repo))
     (:info*
      (lambda ()
        (format
@@ -1317,7 +1317,8 @@ upstream remote."
       (format
        (propertize "Adding %s to database," 'face 'transient-heading)
        (propertize (forge--scope 'url) 'face 'bold)))
-    ("r" forge-forge.remote :format " %k from %d %v," :face 'bold)
+    ("r" forge-forge.remote :if (##forge--scope 'wtree)
+     :format " %k from %d %v," :face 'bold)
     ("a" "pulling all topics"
      (lambda (repo)
        (interactive (list (forge--scope 'repo)))
--- a/lisp/forge-commands.el
+++ b/lisp/forge-commands.el
@@ -378,6 +378,7 @@ commit, and for a file."
       (and$ (magit-file-at-point)            (forge-get-url :blob nil $))
       (forge-post-at-point)
       (forge-current-topic)
+      (forge--loading-topic-url)
       (and (or magit-buffer-file-name
                buffer-file-name
                (derived-mode-p 'dired-mode))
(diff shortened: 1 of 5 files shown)


tarsius:
Was an LLM involved in the creation of these changes?

lucasec:
This change contains no LLM-authored code, and the prose above is my own.

It is in line with the GNU contribution guidelines, which currently prohibit LLM-generated code. I will disclose (as I've noticed that others on the `emacs-devel` mailing list have) that I do use LLMs occasionally to maintain my init, including researching whether certain functionality exists in packages that I have installed.

btw... took me a few mins to find your contribution guidelines again (I know I checked them before submitting this PR). You may want to consider syncing this repo's `CONTRIBUTING.md` with the one in magit: https://github.com/magit/magit/blob/main/.github/CONTRIBUTING.md

lucasec:
I realize this probably should have started as a discussion first, though I had already invested a weekend in a proof-of-concept, so thought it might be easier to just to share the branch (which admittedly took a little bit longer than that to polish up).

tarsius:
I am trying to take a break and this is a lot of code to review, so you will likely have to wait two months or so before I get to it.

lucasec:
Thanks for taking the time to respond. Does pretty much everything magit and forge go single-threaded through you? I assume you have plenty on your plate.

I'm not in a huge hurry to merge this. If it helps, I will also break out the second commit (which is small and relatively self-contained) to its own PR. I assume not much will change about Forge in the interim.

When you do get back, I guess the first question is whether the feature is a good fit for forge. The implementation can always be re-worked.

tarsius:
I've written my own implementation. It adds much fewer new lines.

lucasec:
Your implementation is still asynchronous, right? So if you invoke this command, it may take a second or two to track the repository then fetch the topic before displaying the new buffer?

There is a reason why I added the temporary loading buffer, as otherwise, running the command with prefix commands such as `other-window-prefix` (`C-x 4 4`) or `other-tab-prefix` (`C-x t t`) will not work (the prefix is lost in the asynchronous call chain).

I am okay with the other compromises you made, e.g. skipping the transient to track the repo and always tracking it as selective (this will be fast on github, but slow on gitlab which stubs out the selective fetch to fetch all).

PR #870, opened 2026-08-26
forge-alist: Add forge.fedoraproject.org

mavit:
Assuming you’re accepting additions to the default `forge-alist`, here’s one for `forge.fedoraproject.org`.


tarsius:
Thanks but I think we shouldn't add any project-specific forges to the default value. (I should probably remove the debian entry.)

